Issue #9, opened 2016-11-22
Expected empty line between multiple components


geniusgordon:
I'm getting a **rule-non-nested-empty-line-before** error when I have multiple components in one file

I've got a file like this

```js
import styled from 'styled-component';

const TextA = styled.span`
  color: blue;
`

const TextB = styled.span`
  color: red;
`
```

and got this error
```
7:1  ✖  Expected empty line before non-nested rule   rule-non-nested-empty-line-before
```

I've console.log the extractedCSS and it looks like this
```css
.selector {
  color: blue;
}  /* <----- missing a newline here */
.selector {
  color: red;
}
```

I'm using:
* stylelint-config-standard@15.0.0
* stylelint-processor-styled-components@0.0.4
* stylelint@7.6.0

joggienl:
As a workaround it could help to add an additional rule to your .stylelintrc file:

```
"rules": {
    "rule-non-nested-empty-line-before": "never-multi-line"
  }
```

There are more options for this available. See here: http://stylelint.io/user-guide/rules/rule-non-nested-empty-line-before/

ismay:
@joggienl That doesn't fix it. Only way of getting rid of it for me is to turn it off completely:

```json
{
  "processors": ["stylelint-processor-styled-components"],
  "extends": "stylelint-config-standard",
  "syntax": "scss",
  "rules": {
    "rule-empty-line-before": 0
  }
}
```

chaucerbao:
Maybe `rule-empty-line-before` should be added to the [list of ignored rules](https://github.com/styled-components/stylelint-processor-styled-components/blob/master/src/index.js#L21-L26).

ismay:
@chaucerbao Actually the list of ignored rules should be removed completely. See this pr: https://github.com/styled-components/stylelint-processor-styled-components/pull/23. As I understand it, this lib modifies the Result passed to stylelint, which can lead to swallowed errors that still cause stylelint to exit with an error code: https://github.com/styled-components/stylelint-processor-styled-components/issues/22

A better approach would be to just list the rules that should be ignored for this lib in the readme, so people can add it to their config. Or just create a preconfigured ruleset that people can extend, if you want to make it easy.

ismay:
@mxstbr If you need help with this repo I'd be happy to. Let me know if I can do anything!

rajkuldip:
Below doesn't seems working
` 
"rules": {
    "rule-empty-line-before": 0
  }
`
[...]

PR #11, opened 2016-11-26
Add newline between components' styles

geniusgordon:
#9 

declare extractedCSS as an array, use `.join('\n')` when return

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,7 +48,7 @@ module.exports = (/* options */) => ({
       ],
     })
 
-    let extractedCSS = ''
+    let extractedCSS = []
     let importedNames = {
       default: 'styled',
       css: 'css',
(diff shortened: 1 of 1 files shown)


mxstbr:
Could you add some tests for the new behaviour please? Thank you!

zalmoxisus:
Seems like twice decreasing the perf: https://jsperf.com/string-concat-via-array

geniusgordon:
@mxstbr I don't know how to write tests for this, can't come up with an invalid testcase that produce the **rule-non-nested-empty-line-before** error. Any suggestions? Thank you!

geniusgordon:
@zalmoxisus nice point though, never notice that perf issue

mxstbr:
Sorry for the extremely slow response here, does this fix your issue? If so, why can't we add tests for it?

geniusgordon:
It's ok now! I disabled `rule-empty-line-before` like the comment in #9 said

EnoahNetzach:
I need this fix, so sorry if I opened another PR, but I'll cherry-pick these commits, so attribution should remain untouched.

PR #14, opened 2017-01-23
syntax update for stylelint npm scripts cmd

benkeen:
Using single quotes in the stylelint command didn't work for me (PC, Stylelint 7.7.1). This stackoverflow page simply suggested replacing single quotes with double: http://stackoverflow.com/questions/38814200/files-glob-patterns-specified-did-not-match-any-files - this fixed it for me. Checking out the stylelint docs I didn't see any reference to this issue but all examples used double quotes, so this should be safe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Add a `lint:css` script to your `package.json`. This script will run `stylelint`
 ```JSON
 {
   "scripts": {
-    "lint:css": "stylelint './components/**/*.js'"
+    "lint:css": "stylelint \"./components/**/*.js\""
   }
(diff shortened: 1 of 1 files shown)


mxstbr:
Awesome, thanks!

burcakulug:
@mxstbr , I had the same issue today and I was trying to figure it out. This fix also worked for me, but someone changed the documentation back to single quotes, so I didn't see this in the README on master when I was first setting it up.

burcakulug:
No quotes also worked, I guess it is okay if there is not space.

Issue #13, opened 2017-01-23
Unexpected unknown property (property-no-unknown)


shripadk:
Not sure if this is related to #6 (please close if so) but it breaks with the following complex interpolation:

```javascript
  const Button = styled.button`
    margin-${props => ((props.theme.dir === 'rtl') ? 'left' : 'right') }: 12.5px;
  `;
```

This is used for providing RTL support where the **dir** property can be set in `ThemeProvider` like so:

```javascript
<ThemeProvider theme={{ dir: 'rtl' }}>
  <Button>Hello, World!</Button>
</ThemeProvider>
```

Stylelint throws the following error: _Unexpected unknown property "margin-$undefined" (property-no-unknown)_.

mxstbr:
Ah interesting, that's a bug that'll be interesting to fix! Mind submitting a PR with a failing test case? That'd be super helpful.

shripadk:
Have submitted a PR with failing test cases. Let me know if you need anything else :)

emilgoldsmith:
This should be able to be handled with an `sc-custom` tag as added in #70 when we publish it in the near future.

PR #17, opened 2017-02-09
Fixed complex interpolations

### LINKED ISSUE #6: Breaks on complex interpolations
Currently, this breaks on a complex interpolation:

``` JS
const background = 'background: red;';
const Button = styled.div`
  color: red;
  ${background}
`;
```

We represent this for stylelint like so:

``` CSS
.selector {
  color: red;
  $background
}
```

The problem is that, while stylelint ignores Sass variables, those cannot represent full strings, which means stylelint throws a "CssSyntaxError" on the above code eventhough it's not a syntax error at all. 😕

Any ideas what we could do here @davidtheclark?

(also related to #4)

rainydio:
Merged tests from #8 and fixed #6. This change complicates `interleave` function and also introduces *vendor prefixed* property `-styled-mixin`.

```js
const Button = styled.button`
  color: blue;
  ${additionalButtonStyles}
`
```

Is now transformed to:
```css
.selector {
  color: blue;
  -styled-mixin: additionalButtonStyles;
}
```

### DIFF
--- a/src/utils/general.js
+++ b/src/utils/general.js
@@ -29,9 +29,26 @@ const fixIndentation = (str) => {
   }
 }
 
+/**
+ * Checks if last line of text has whitespaces only
+ */
+const isLastLineWhitespaceOnly = (text) => {
+  for (let i = text.length - 1; i >= 0; i -= 1) {
+    const char = text.charAt(i)
+    if (char === '\n') {
+      return true
+    }
+    if (char !== '\t' && char !== ' ') {
+      return false
+    }
+  }
+  return true
+}
+
 const wrapSelector = (content) => `.selector {${content}}\n`
 const wrapKeyframes = (content) => `@keyframes {${content}}\n`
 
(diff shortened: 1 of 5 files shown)


rainydio:
It doesn't fix #13 but there is workaround now. Instead of:
```js
const Button = styled.button`
    margin-${props => ((props.theme.dir === 'rtl') ? 'left' : 'right') }: 12.5px;
`;
```
write:
```js
const Button = styled.button`
    ${props => props.theme.dir === 'rtl'
        ? 'margin-left: 12.5px'
        : 'margin-right: 12.5px'
    }
`;
```

chuanxie:
@rainydio  how to solve the media query?
like
```
const aStyled = styled.a`
    ${media.tablet`display: none;`};
`
```
any possible workaround? `-styled-mixin` doesn't work here

HejdesGit:
Any progress with this? Something missing from accepting the PR?

PR #23, opened 2017-04-15
Fixed: set `errored` to `false` if `warning` no contains errors.

### LINKED ISSUE #22: Incorrectly exits with exit code 2 for only warnings
Stylelint cli exits with exit code 2 for a run with only warnings (see [this build](https://travis-ci.org/ismay/ismaywolff.nl/builds/219376502#L560)) when using stylelint-processor-styled-components. After removing stylelint-processor-styled-components from the build chain, the process exits with 0, as expected.

Since removing the processing fixes the issue I'm suspecting that this lib is the cause, but I'm unsure where exactly it's going wrong. Is it possible that this processor somehow could cause the process to exit with an error code of 2 instead of 0?

evilebottnawi:
Closes #22

BTW, i would recommend don't changed `stylelintResult`, just add node in `README.md` about disable `no-empty-source` and `no-missing-end-of-source-newline`, in next version we can change `stylelintResult` (example rename `warnings` to another name) and `stylelint-processor-styled-components` will cease to work as expected.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -95,8 +95,9 @@ module.exports = (/* options */) => ({
       prevWarnings.push(correctedWarning)
       return prevWarnings
     }, [])
-    
+
     if (newWarnings.length === 0) {
+      // eslint-disable-next-line no-param-reassign
       stylelintResult.errored = false
     }
 
--- a/src/index.js
+++ b/src/index.js
@@ -95,6 +95,10 @@ module.exports = (/* options */) => ({
       prevWarnings.push(correctedWarning)
       return prevWarnings
     }, [])
+    
+    if (newWarnings.length === 0) {
+      stylelintResult.errored = false
+    }
 
     return Object.assign(stylelintResult, { warnings: newWarnings })
   },


mxstbr:
> in next version we can change stylelintResult (example rename warnings to another name) and stylelint-processor-styled-components will cease to work as expected.

That would be a major version release anyway, in which case we'd have to update and release a major version too. Might as well make that explicit rather than only mentioning it in the docs and avoid a footgun for people?

mxstbr:
Can you fix the linting errors?

```
   98:1  error  Trailing spaces not allowed                                     no-trailing-spaces
  100:7  error  Assignment to property of function parameter 'stylelintResult'  no-param-reassign
```

ismay:
> BTW, i would recommend don't changed stylelintResult

@evilebottnawi Next to this pr as a patch, I think it would be a good solution to define the rules that need to be disabled in a shareable config. Users can then extend that config and list it before all others so it overrides all other rules (https://stylelint.io/user-guide/configuration/#extends).

Say for example: `stylelint-config-styled-components`

Then we could leave `stylelintResult` as is. I think defining the shareable config as a peerDependency could be a good way to allow coupling a version of this lib to a version of the config. What do you think?

Also, would you mind fixing the linting errors? If you don't have time let me know. If you could then allow me access to the pr I'd be happy to help.

evilebottnawi:
@ismay sorry, no time now for this, maybe late :disappointed:

ismay:
> sorry, no time now for this, maybe late

No worries, linting errors are fixed 👍

ismay:
I'll try and rebase it this weekend btw, so we can get this merged.

PR #28, opened 2017-05-18
use unique name for each wrapped selector

ramitos:


mxstbr:
Is this necessary/does it fix anything? I haven't run into any trouble just using `.selector`...

ramitos:
yes, it conflicts with https://github.com/primer/stylelint-config-primer/blob/master/index.js#L263

ramitos:
@mxstbr incrementing selector is a simple and great idea, will do that!

ramitos:
@mxstbr updated

mxstbr:
This is great, thanks so much!

PR #33, opened 2017-06-12
Move typescript to dependencies

### LINKED ISSUE #32: Requires typescript in non typescript project
Running this plugin on a non typescript problem yielded the following error

```
Error: Cannot find module 'typescript'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/node_modules/typescript-eslint-parser/parser.js:11:10)
```

Installing typescript fixes the problem but it should not be required on a non typescript project.

ismay:
See #32. This pr fixes having to install typescript manually (typescript-eslint-parser requires it as a peerdependency). However it does mean that there's no more freedom for the user to choose which exact version they want to use. But since typescript-eslint-parser has a specific dependency range (currently `~2.3.2`) the choice is limited anyway.

Closes #32

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,12 +61,6 @@
       "integrity": "sha1-MU3QpLM2j609/NxU7eYXG4htrzw=",
       "dev": true
     },
-    "align-text": {
-      "version": "0.1.4",
-      "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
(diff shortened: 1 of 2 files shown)


ismay:
@mxstbr Could you release this on npm? Patch release would be appropriate here I think. I wouldn't mind doing it either, but I don't have the rights. If you want to add me I'm @ismay on npm as well (https://www.npmjs.com/~ismay)

PR #21, opened 2017-04-04
fixture for consecutive template literals

TimeRaider:
related to #20

Issue #20, opened 2017-04-04
Incorrectly fails on 2 or more consecutive template literals


TimeRaider:
```js
import styled from 'styled-components';
const gap = '5px';
const Button = styled.div`
    padding: 0 ${gap} ${gap};
`
```
produces
`Missed semicolon   CssSyntaxError`

nathggns:
+1

ismay:
Closed in #52

PR #52, opened 2017-06-27
Fixed multiple interpolations in a property

taneba:
related to #20 #21

This change enable to use multiple interpolations in a property at once.

```
  width: 20px;
  border: ${borderWidth} ${borderStyle} ${color};
```
is now translated to:

```
 width: 20px;
 border: $borderWidth -styled-mixin: borderStyle; -styled-mixin: color;
```
This changes above to: 

```
 width: 20px;
 border: $borderWidth $borderStyle $color;
```

Cc: @mxstbr

### DIFF
--- a/test/utils.test.js
+++ b/test/utils.test.js
@@ -41,15 +41,15 @@ describe('utils', () => {
 
   describe('isLastLineWhitespaceOnly', () => {
     it('should return true for empty string', () => {
-      expect(isLastLineWhitespaceOnly('')).toEqual(true)
+      expect(isLastLineWhitespaceOnly('')).toEqual(false)
     })
 
     it('should return true for string of spaces', () => {
-      expect(isLastLineWhitespaceOnly('   ')).toEqual(true)
+      expect(isLastLineWhitespaceOnly('   ')).toEqual(false)
     })
 
     it('should return true for string of spaces and tabs', () => {
-      expect(isLastLineWhitespaceOnly(' \t  ')).toEqual(true)
+      expect(isLastLineWhitespaceOnly(' \t  ')).toEqual(false)
     })
 
     it('should return false for string with something other than space and tab', () => {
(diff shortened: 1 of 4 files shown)


test/utils.test.js
```diff
@@ -41,15 +41,15 @@ describe('utils', () => {
 
   describe('isLastLineWhitespaceOnly', () => {
     it('should return true for empty string', () => {
-      expect(isLastLineWhitespaceOnly('')).toEqual(true)
+      expect(isLastLineWhitespaceOnly('')).toEqual(false)
```
mxstbr: I don't quite get why this is false now? That line is only whitespace? (same for all other lines)

taneba: I think isLastLineWhitespaceOnly() basically resolves the interpolation which includes   property and value.

```
`${props => props.primary && css`background: #ffffff;`}`
```
in this case, isLastLineWhitespaceOnly returns true and then interleave() provides `-styled-mixin: undefined;`.

And when interpolation is value of any property: 
```
`background: ${color}`
```
in this case, isLastLineWhitespaceOnly returns false and then interleave() provides `background: $color;`.



But what if the multi interpolations in value?

```
`background: ${box} ${color}`
```
In this case, isLastLineWhitespaceOnly now returns true and then interleave() provides `background: $box -styled-mixin: color;` that causes syntax error.
It is because quasis[1] is `' '`(between two interpolations).

So I change isLastLineWhitespaceOnly returns false when receive `' '` or `''` , then above case returns true and interleave() provides `background: $box $color`. 


Is this correct?

ismay: I don't really get how this fixes the interpolations either. If it does that's great but as I see it `isLastLineWhitespaceOnly` now returns the exact opposite of what I'd expect it to return.

I'd suggest refactoring some more so that the cause of the interpolations breaking is fixed without breaking isLastLineWhitespaceOnly's current behaviour, or at least keeping the behaviour in line with the function name.

taneba:
I've refactored by following suggestion by @ismay 
How about this?

ismay:
Thanks for the update @taneba. To me this looks good. If there aren't any objections I think this is ready for merging.

ismay:
Thanks for the PR @taneba !

PR #57, opened 2017-07-06
Add stylelint-disable note in README.md

emilgoldsmith:
I was trying to think of a better example for a rule to break, but as I'm new to stylelint I couldn't find a good one yet, if anyone has a better example feel free to suggest that.

Especially using the media queries (though it is an example in the docs) could be a bit confusing I guess, but yeah, as I'm new to stylelint it was just hard to think of good examples.

It'd be neat to have in here though so no one else submits erroneous issues like I did :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,23 +57,17 @@ npm run lint:css
 ```
 
 ### Syntax notes
-#### stylelint-disable
-In order for `stylelint-processor-styled-components` to parse your `stylelint-disable` comments they must be inside the actual Styled Components CSS as such:
+#### Turning rules off from within your CSS
+In order for `stylelint-processor-styled-components` to parse your `stylelint-disable` comments (see the [stylelint documentation](https://stylelint.io/user-guide/configuration/#turning-rules-off-from-within-your-css) for all allowed syntax) they must be inside the actual Styled Components CSS as such:
 
 **Wrong**:
 ```
-/* stylelint-disable declaration-empty-line-before */
+/* stylelint-disable color-named */
 import React from 'react';
 import styled from 'styled-components';
 
-const media = ...
-
 const Wrapper = styled.div`
   background-color: red;
-  
-  ${media.desktop`background-color: blue;`}
-  ${media.tablet`background-color: yellow;`}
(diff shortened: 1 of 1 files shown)


ismay:
> Especially using the media queries (though it is an example in the docs) could be a bit confusing I guess, but yeah, as I'm new to stylelint it was just hard to think of good examples.

Yeah you're right, I think the media queries might make it too complicated. You could just do this:

**Wrong**
```javascript
/* stylelint-disable color-named */
import React from 'react';

const Wrapper = styled.div`
  background-color: red;
`;
```

**Right**
```javascript
import React from 'react';

const Wrapper = styled.div`
  /* stylelint-disable color-named */
  background-color: red;
`;
```

Also stylelint supports different types of configuration comments from within the css. So it's probably best to change the  'stylelint-disable' heading to 'Turning rules off from within your CSS', as that's what it's called in the stylelint docs. Also it'd be good to add a link to this section of the stylelint docs: https://stylelint.io/user-guide/configuration/#turning-rules-off-from-within-your-css

> It'd be neat to have in here though so no one else submits erroneous issues like I did :).

No worries, issues with the docs are also valid issues! And prs like this are definitely appreciated 👍

emilgoldsmith:
Just committed changes. I really like your example, both very simple and seems pretty realistic at the same time!

Seems the URL hash doesn't work in my browser anyway which is a bit sad, but people will just have to figure that out I guess.

And yeah maybe I'll find a bit of time outside of work to help with a few of the other issues, especially the ones we encounter while picking up Styled Components for our MVP here at my startup [Gnarles](www.gnarles.com) :).

ismay:
> Seems the URL hash doesn't work in my browser anyway which is a bit sad, but people will just have to figure that out I guess.

Yeah same here.

> And yeah maybe I'll find a bit of time outside of work to help with a few of the other issues, especially the ones we encounter while picking up Styled Components for our MVP here at my startup Gnarles :).

That would be awesome!

ismay:
Thanks @emilgoldsmith for the PR!

emilgoldsmith:
No problem @ismay, happy to help!

PR #59, opened 2017-07-06
Make -styled-mixins unique and clarify interpolation linting in readme

### LINKED ISSUE #55: Error declaration-block-no-duplicate-properties from rules with mixins only
My code:
```
import { ellipsis } from 'polished';
import { TYPOGRAPHY } from 'style-constants';

const NameLink = styled.span`
  ${ellipsis('110px')}
  ${TYPOGRAPHY.bodyBold}
`;

export default NameLink;
```

 7:3  ✖  Unexpected duplicate "undefined"   declaration-block-no-duplicate-properties

emilgoldsmith:
Up until now I just added a failing test for #55. It seems to be a problem with [this line](https://github.com/styled-components/stylelint-processor-styled-components/blob/3c43e97f5678ece590cd6a284b0e943fc69073af/src/utils/tagged-template-literal.js#L21) returning undefined on a string as opposed to a variable.

I'll probably look at it again tomorrow night after work, and see if I can also fix it, my first impression is that it's a bigger problem with the approach and that it's better to find a better approach rather than patching an edge case for another one to soon pop out, but again, I haven't looked at it that much yet.

Any advice from more seasoned contributors that know the codebase better is of course welcome, otherwise I'll look at it again tomorrow night.

(fixes #55)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,8 @@ const Wrapper = styled.div`
 `;
 ```
 
+#### Interpolation linting
+We do not currently support linting interpolations as it could be a big performance hit though we aspire to have at least partial support in the future. You can of course lint your own mixins in their separate files, but it won't be linted in context, the implementation currently just inserts relevant dummy values. This, we are afraid, means you won't be able to lint cases such as `declaration-block-no-duplicate-properties` etc. and won't be able to lint outside mixins such as [polished](https://github.com/styled-components/polished).
 
 ## License
 
--- a/src/utils/tagged-template-literal.js
+++ b/src/utils/tagged-template-literal.js
@@ -14,21 +14,23 @@ const hasInterpolations = node => !node.quasi.quasis[0].tail
 /**
  * Merges the interpolations in a parsed tagged template literals with the strings
  */
+// Used for making sure our dummy mixins are all unique
+let count = 0
 const interleave = (quasis, expressions) => {
   let css = ''
   for (let i = 0, l = expressions.length; i < l; i += 1) {
     const prevText = quasis[i].value.raw
     const nextText = quasis[i + 1].value.raw
-    const expression = expressions[i]
 
     css += prevText
     if (isLastLineWhitespaceOnly(prevText) && !isEmptyOrSpaceOnly(prevText)) {
-      css += `-styled-mixin: ${expression.name}`
+      css += `-styled-mixin${count}: dummyValue`
+      count += 1
       if (nextText.charAt(0) !== ';') {
         css += ';'
       }
     } else {
-      css += `$${expression.name}`
+      css += '$dummyValue'
     }
   }
   css += quasis[quasis.length - 1].value.raw
--- a/test/fixtures/interpolations/valid.js
+++ b/test/fixtures/interpolations/valid.js
@@ -42,6 +42,7 @@ const Box3 = styled(Box2)`
 `
 
 // Multiline
+// prettier-ignore
 const Button4 = styled.button`
   display: block;
   ${`
@@ -52,6 +53,7 @@ const Button4 = styled.button`
 
 // Conditional
 const cond = true
+// prettier-ignore
 const Button5 = styled.button`
   display: block;
   ${cond &&
@@ -63,10 +65,11 @@ const Button5 = styled.button`
 
 // Conditional
 const cond2 = false
+// prettier-ignore
 const Button6 = styled.button`
   display: block;
   ${cond2 &&
-    `
+   `
     color: ${cond2};
   `}
   background: blue;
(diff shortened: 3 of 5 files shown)


emilgoldsmith:
Hmm... What's up with the oneliner and the random line collections in my commit? Is that what husky was doing with the pre-commit stuff? Definitely wasn't what I was doing :P. How can I avoid that?

emilgoldsmith:
Just had a bit of a think about it on the train to work again.

My first thought for a quickfix would be doing this in `tagged-template-literal.js`:
```
let count = 0;
/**
 * Merges the interpolations in a parsed tagged template literals with the strings
 */
const interleave = (quasis, expressions) => {
  let css = ''
  for (let i = 0, l = expressions.length; i < l; i += 1) {
    const prevText = quasis[i].value.raw
    const nextText = quasis[i + 1].value.raw
    const expression = expressions[i]
    const name = expression.name || `dummy${count++}`

    css += prevText
    if (isLastLineWhitespaceOnly(prevText)) {
      css += `-styled-mixin: ${name}`
      if (nextText.charAt(0) !== ';') {
        css += ';'
      }
    } else {
      css += `$${name}`
    }
  }
  css += quasis[quasis.length - 1].value.raw
  return css
}
```

It's a bit of a quickfix but I guess it keeps your current approach alive?

I'm guessing there isn't any smart library that could handle actually evaluating the interpolations for us? We couldn't evaluate the strings first with the actual styled library or something? Would of course be awesome but I can see it could also be a big amount of work as you'd have to parse imports etc.

ismay:
> Hmm... What's up with the oneliner and the random line collections in my commit? Is that what husky was doing with the pre-commit stuff? Definitely wasn't what I was doing :P. How can I avoid that?

That's probably [prettier](https://github.com/prettier/prettier). Looks like a bug.

ismay:
> my first impression is that it's a bigger problem with the approach and that it's better to find a better approach rather than patching an edge case for another one to soon pop out, but again, I haven't looked at it that much yet.

I think so too. For now I haven't had enough time to really consider which approach would be the most maintainble, but I'm also expecting that a slight refactor to clean things up a bit would be good. I'm intending to make things a bit more modular to prevent things from getting too complex.

> I'm guessing there isn't any smart library that could handle actually evaluating the interpolations for us? We couldn't evaluate the strings first with the actual styled library or something? Would of course be awesome but I can see it could also be a big amount of work as you'd have to parse imports etc.

That might be possible in some cases, but ultimately I think it would be quite difficult. Some values would only be available at runtime if I'm not mistaken and there would probably be quite a performance hit.

emilgoldsmith:
>That might be possible in some cases, but ultimately I think it would be quite difficult. Some values would only be available at runtime if I'm not mistaken and there would probably be quite a performance hit.

Yeah I agree to that. Could definitely make a big performance hit. It is a bit sad you can't have your interpolations linted as of right now though, I might try looking into if there is some smart library we could use to handle some of the cases as I believe most cases should be known at compile time, and maybe also looking into different approaches and get back to you here.

ismay:
> Yeah I agree to that. Could definitely make a big performance hit. It is a bit sad you can't have your interpolations linted as of right now though, I might try looking into if there is some smart library we could use to handle some of the cases as I believe most cases should be known at compile time, and maybe also looking into different approaches and get back to you here.

Great! Thanks for taking a look at it, appreciate the help 👍

emilgoldsmith:
> edit: Might be related to prettier/prettier#2291

Yeah seems very likely this is the problem. I'll add some `// prettier-ignore` statements.

> Great! Thanks for taking a look at it, appreciate the help :+1:

No problem! I already took a bit of time to look into it now actually, and I'll be pushing up my idea of a fix + commenting my reasoning soon :).

emilgoldsmith:
Okay, a few things:

1. Force pushed the first commit to add `// prettier-ignore` statements and also rebased onto newly pushed code from #52 etc.

2. I looked into actually parsing a bit, and also just thought about it. While I believe in practice we could probably solve many cases as mostly people might just import mixins from a single file that they store it in / substitute variables declared in that same file, the worst case is of course that we'd have to emulate a full webpack build basically which would of course be horrendous for performance. And even with the things we could possibly support, it would still be a bigger project, so if you're okay with it @ismay and maybe if you have an issue tag for "long term goals" or something of equivalent meaning, I'll open an issue on that though it is probably low priority. It could also possibly be something that could be a goal for my webpack linter.
To reflect the "giving up" on that I added a piece to the README explaining that we don't support interpolation linting so people won't be surprised by that.

3. My solution which I think is good enough for now, especially given that we add a piece to the README is just making all the -styled-mixins unique by adding a count. And then I just changed the `expression.name` to `dummyValue` because I think that reflects the actual use of it better? I mean, since the `expression.name` wasn't really being used for anything other than a placeholder to make stylelint pass over it quietly, right? If I missed a usecase for the `expression.name` please say so, but I also thought since it returns `undefined` on anything that's not a variable I'm assuming it didn't have any other uses.

Thoughts? :)

emilgoldsmith:
Also, from my POV this is ready to merge as of now, pending your comments.

mxstbr:
Great work! 👌

emilgoldsmith:
Thanks! And thanks for the invite :). Look forward to working with you in my spare time :).

mxstbr:
Thanks for the great work so far, happy to have you on board!

PR #65, opened 2017-07-08
README Coveralls url was erroneously linking to a branch called ci

emilgoldsmith:
I guess it was just a silly mistake when linking up coveralls while being on the wrong branch


emilgoldsmith:
No problem! Happy to help :). And thank you for innovating in the right direction :).

PR #16, opened 2017-01-30
Add failing tests for #13

shripadk:
- Failing test case for valid interpolations
- Real world failing test case (BidiButton)

Issue #68, opened 2017-07-10
How to fix all these interpolation edge cases?


emilgoldsmith:
(first of all, hope it's all good I added a discussion label)

So, there are obviously heaps of interpolation edge cases, I think all issues labeled with `bug` other than #26 are currently due to this, and I listed 4 cases (some of them overlapping with already existing issues) in  the comments of #63. So I think we should have a bit of a discussion about how to fix this if you're up for it @styled-components/stylelint-processor. 

I see the biggest issues with this as:
- Many things will simply be unknowable before runtime such as what the props will be to a component etc.
- If we were to try doing some smart stuff with interpreting Javascript to figure out the parts we could from pre-runtime it would probably take a lot of man-hours and could also potentially really damage our performance in quite a few cases.
- Currently we are enumerating cases based on assumptions of use, and it's always a difficult path to for a proof (yep, I think like a mathematician :stuck_out_tongue:), and therefore for finding a bug-free missed-edge-case-free approach as well.

An idea came to me while pondering about the different things I've been working on in this repo recently, that should definitely work, and not be too hard to implement. I also don't think it would hinder UX too badly, but that's definitely up for discussion. It's definitely not a dream-worthy perfect solution as the user would have to do a bit of work, but let's see what you think, and otherwise that's also why the discussion tag is here!

So my approach would:
- Ditch any attempts at trying to interpret JS, and continue on the path of inserting dummy values that will stop Stylelint from complaining, and then interpolations will either be linted where it was defined if defined with a `styled-components` helper, or not at all. It will hopefully do it in a much more comprehensive and correct way though, so we could close all those `bug` labeled issues all at once :grinning: 
- Make the user declare what kind of output this interpolation will have at the start of every interpolation in a comment, and we can then easily substitute the right dummy value
- In case the user doesn't declare the type we can keep defaults such as something what we're using now, and also of course let that be known, so you would only have to declare in special cases and we can hopefully guess for the rest like we are doing now.

So implementation wise I would do something like this:
- I would plan on using something like [this css vocabulary list](http://apps.workflower.fi/vocabs/css/en) and not needing people to type it all out, but maybe make some shortcuts and if a substring has a unique match then use it etc. maybe store it as a trie or something. This would make sure we covered basically all edge cases, and the few that would possibly arise after should not be so numerous that it would feel bad adding a few more special-cases.
[...]

PR #70, opened 2017-07-14
The feature to fix 'em all!

### LINKED ISSUE #13: Unexpected unknown property (property-no-unknown)

### LINKED ISSUE #34: Doesn't properly handle contextual selectors
Causes errors to trigger when using contextual selectors like `${Button}`

ie. 
```
export const ButtonGroup = styled.div`
  white-space: nowrap;

  > ${Button} {
    margin: 0;
    border-radius: 0;
    vertical-align: top;
  }
`
```

```
149:28  ✖  Expected "$Button" to be "$button"     selector-type-case
 149:28  ✖  Unexpected unknown type selector       selector-type-no-unknown
            "$Button"
```

### LINKED ISSUE #54: Strange errors with refering to other components
```
import { div } from 'react-dom';

const IconWrapper = styled.span`
  display: inline-block;
  ${div} & {
    color: #fff;
  }
`;

export default IconWrapper;
```

```
app/components/LeftMenu/IconWrapper.js
 6:22  ✖  Expected newline after ";" in a multi-line declaration block   declaration-block-semicolon-newline-after
 6:23  ✖  Expected empty line before rule                                rule-empty-line-before
```

emilgoldsmith:
This feature should provide the tools to close #54, fix #34, and resolve #13.

The feature to be implemented in this PR is the one discussed in #68. And I will open another discussion issue later to talk about how the final API should look like and edit it in here.

The main points needed to be implemented for the full feature is:
- [x] actually parsing the comments
- [x] finalize the API (decide which labels to include and how to substitute for each of them)
- [x] connect parsing and the API
- [x] write thorough tests including the tests added in #16 (and therefore close #16)
- [x] Document everything thoroughly in the README


emilgoldsmith:
This is currently just a dummy PR to outline the main points needed to be implemented. I may or may not find time to actually work on it this weekend, I'll probably finalize #67 first, and I have a few other things I want to work on as well. 
Opening this PR also allows other people to take a stab at implementing it if they feel like it, it's definitely on my to-do list though but anyone is free to beat me to it :).

emilgoldsmith:
Proof of concept for the shortened commands + probably even the final implementation of it implemented @mxstbr @ismay. It would be way overkill using a trie or something with better complexity when we'll only have 10 commands or something like that to check against, so I think this should work just fine, right? :)

emilgoldsmith:
Okay, we're nearly there!

I just need a tiny bit of feedback, specifically concerning the handling of whitespacee in the comments of #87, and then I'd also like to hear your opinion on what we should do if someone writes an invalid tag, as for example `/* sc-invald */`, should we throw an error and let the user know they errored, or should we just ignore it and implement our default, silently failing?

I think I prefer the first with letting them know as the fact they prefix with `sc-` must mean they want to use our API and in that case I feel like we should let them know if they made a typo etc. which would create unexpected behaviour.

Other than that I just need to polish a bit of code, polish / add a few more tests, and update the README.
I'm assuming I'll just add the docs to our README initially and then we can move it to the website soon after, yeah?

ismay:
> what we should do if someone writes an invalid tag, as for example /* sc-invald */, should we throw an error and let the user know they errored, or should we just ignore it and implement our default, silently failing?

Hmm, not sure. As in an actual error or a stylelint error?

emilgoldsmith:
> Hmm, not sure. As in an actual error or a stylelint error?
Yeah that's also a relevant question which my initial thought at least was to throw a real error, because I thought it was a bit weird making custom stylelint errors, but maybe that could actually work I guess. We could call the rule `styled-components/interpolation-tag` or something like that

ismay:
If you make it a custom stylelint rule it'd be something we could probably neatly solve by including it in stylelint-config-styled-components-processor

emilgoldsmith:
> If you make it a custom stylelint rule it'd be something we could probably neatly solve by including it in stylelint-config-styled-components-processor

Yeah that'd be one way. I feel like it would create a bit of a dangerous possible mismatch between versions though, it would be a very strong peerDependency anyway in terms of if the API changes we need to change the rule.
That'd also mean that the rule wasn't static which I guess is suboptimal.

The other way I thought was to maybe to inject a stylelint error if we went down the stylelint error path. I'm not sure exactly how you would inject it but I'd hope it was possible before the final part where we were previously removing errors from the result (before we had the config repo), though that would of course definitely be possible.

ismay:
> I feel like it would create a bit of a dangerous possible mismatch between versions though, it would be a very strong peerDependency anyway in terms of if the API changes we need to change the rule.

That's true. Although in the worst case scenario it'd be false positives, or it not warning when it should (if you allow the processor to just fail silently).

> The other way I thought was to maybe to inject a stylelint error if we went down the stylelint error path. I'm not sure exactly how you would inject it but I'd hope it was possible before the final part where we were previously removing errors from the result (before we had the config repo), though that would of course definitely be possible.

That should be possible. But I'd prefer not to rely on the internal structure of the stylelintResult. I'd say that if we want it to be reported through stylelint we should do it with a rule. But yeah, that also has its downsides (as you say, the rule could change for example).

Another way would be to throw an error from our processor, but I don't know how stylelint would handle that (if it would crash, or something else).

emilgoldsmith:
> Another way would be to throw an error from our processor, but I don't know how stylelint would handle that (if it would crash, or something else).

Yeah that was my initial idea, and I assumed it would crash, which as I said:
> I think I prefer the first with letting them know as the fact they prefix with sc- must mean they want to use our API and in that case I feel like we should let them know if they made a typo etc. which would create unexpected behaviour.

And I therefore thought that justified crashing stylelint? But yeah, as we both agree there are ups and downs of the different choices, and throwing an error I guess could also do something else than crash stylelint? Though I would be surprised and can't imagine what it would do then.

ismay:
Hmm yeah. I think all things considered I'd then also prefer throwing an error from our processor.

One thing I am worried about is whether editors with stylelint linters will restart stylelint when it crashes, but that maybe isn't really our concern.

It seems like the best first option, and we could always adapt it should it not work as expected.

mxstbr:
Can't wait to land this! Let's put docs into the README for now, then write and publish an article, release 1.0 sometime this week and then work on moving the docs?

emilgoldsmith:
> It seems like the best first option, and we could always adapt it should it not work as expected.

Yeah I think that's probably the main point. Let's start out with throwing an error, and if it screws with editors (which is a very valid point) we can adapt it. Maybe use the rule or something of the like.

emilgoldsmith:
> Can't wait to land this! Let's put docs into the README for now, then write and publish an article, release 1.0 sometime this week and then work on moving the docs?

Sounds like a plan :).

The final feedback I need from you guys is just regarding my last comment in #87, as I think we settled the error thing for now. And then I'll try and make some time in the next few days to finish up this PR :).

emilgoldsmith:
Woot woot, I think I'm done! I'm pretty happy with it now :). I'll just work on the README now and maybe do a bit of squashing of commits for nicer history, and I'm ready for review and merge to our v1.0 branch :).

emilgoldsmith:
Okay! I'm ready for a good bout of review :). Probably mostly looking for feedback on the wording of the README as I think it's okay but could probably use a few tweaks here and there and things always make more sense to you than to outsiders so it's good to have someone else read it over.

And then a small thing is the more semantic question of whether or not to include `sc-ref` in the API spec as @mxstbr had also earlier questioned. I'm a bit in the middle on it without too strong of an opinion as I see arguments on both sides, so maybe the two of you ( @ismay  and @mxstbr ) can help me make a decision? :).

Look forward to merging this in!

README.md
```diff
@@ -112,7 +112,46 @@ would throw a stylelint error similar to `All rules have already been disabled (
 
 #### Interpolation linting
 
-We do not currently support linting interpolations as it could be a big performance hit though we aspire to have at least partial support in the future. You can of course lint your own mixins in their separate files, but it won't be linted in context, the implementation currently just inserts relevant dummy values. This, we are afraid, means you won't be able to lint cases such as `declaration-block-no-duplicate-properties` etc. and won't be able to lint outside mixins such as [polished](https://github.com/styled-components/polished).
+We are afraid that we don't support linting interpolations (meaning the actual CSS that would be inserted into an interpolation through a variable or function during runtime) as it could be a big performance hit. You can of course lint your own mixins in their separate files, but it won't be linted in context, our implementation simply inserts relevant dummy values. This, we are afraid, means you won't be able to lint cases such as `declaration-block-no-duplicate-properties` etc. and won't be able to lint outside mixins such as [polished](https://github.com/styled-components/polished) in context.
```
mxstbr: I would get rid of this paragraph since it's unnecessary, the following paragraphs explain how it works anyway.

mxstbr:
Rewriting the README on this flight, sit tight. Will push to this branch once I land!

mxstbr:
Did a couple of passes of rewrites over the README, let me know if that's alright, and if so let's merge this PR and start prepping the 1.0 release blog post!

emilgoldsmith:
I'm going out right now but I'll look over it tonight :).

So shall we just leave the ref in the API? I'm also okay with that as mentioned earlier

mxstbr:
Yeah I think so, I don't think it's necessary. We can always add it later!

README.md
```diff
@@ -41,90 +43,157 @@ Add a `.stylelintrc` file to the root of your project:
...
+#### Webpack
+
+If you want to lint on build, rather than as a separate command, you can use the [`stylelint-custom-processor-loader`](https://github.com/emilgoldsmith/stylelint-custom-processor-loader) for webpack.
+
 ### Processor specific stylelint rules
 
-When using this processor a couple of stylelint rules throw errors that you cannot prevent. Like
-'[no-empty-source](https://stylelint.io/user-guide/rules/no-empty-source)' or
-'[no-missing-end-of-source-newline](https://stylelint.io/user-guide/rules/no-missing-end-of-source-newline)'.
+When using this processor a couple of stylelint rules throw errors that cannot be prevented, like [`no-empty-source`](https://stylelint.io/user-guide/rules/no-empty-source) or [`no-missing-end-of-source-newline`](https://stylelint.io/user-guide/rules/no-missing-end-of-source-newline). There's also a couple unnecessary ones like [`no-vendor-prefix` rules](https://stylelint.io/user-guide/rules/property-no-vendor-prefix). (`styled-components` automatically vendor prefixes your code, so you don't need to do it manually)
```
emilgoldsmith: You say there are a couple of 'unnecessary rules' here, don't you mean that they're actually necessary? As in, vendor prefixing is unnecessary and therefore we enforce that stylelint rule in our config due to the nature of Styled Components? I think that sentence could use a bit of rephrasing.

README.md
```diff
@@ -41,90 +43,157 @@ Add a `.stylelintrc` file to the root of your project:
...
 
+```js
+const something = 'background';
+
+const Button = styled.div`
+  ${something}: papayawhip;
+`
+```
+
+> **NOTE:** This example is just for illustration purposes, it would lint correctly. Imagine an imported utility from a third-party library, a global variable, etc.
```
emilgoldsmith: I believe this example would actually exactly throw an error, since we would guess it was a whole declaration as this is the start of a new declaration. Therefore with our substitution the CSS would probably end up looking like: 
```
  -styled-mixin0: dummyValue;: papayawhip
``` 
as far as I believe, so this note is incorrect, the example is good though :)

README.md
mxstbr: Sure, feel free to rephrase!

README.md
mxstbr: Perfect, mind removing the note?

mxstbr:
I've also realised that the automatic-shorthand is going to bite us in the ass with versioning as it is implemented right now.

Imagine right now somebody wants to use the `sc-selector` tag. They think to themselves "Hmm, no other tag starts with `sc-s`, so I'll just use that to be short!". They write code like this:

```JS
const Wrapper = styled.div`
  ${/*sc-s*/Button} {
    color: blue;
  }
`
```

Now imagine we want to introduce a `sc-soda` tag. (stupid example, bear with me) We're fucked, because that is now a semver-major change, otherwise the person upgrades and their code errors out. Basically, any tag addition is immediately a semver-major change if the tag starts with anything that's unique right now but won't be then.

Do we plan on introducing additional tags like, ever?

emilgoldsmith:
Okay and I also just now pushed a commit removing `ref` from the API. I also like the way you handled it in the README @mxstbr which makes it very clear and the `sc-ref` much more superfluous

emilgoldsmith:
> Now imagine we want to introduce a sc-soda tag. (stupid example, bear with me) We're fucked

Yeah that's a really good though. Hmmm... I like the shortening a lot though. One way is of course to decide what the shorthand is instead of the dynamic / variable approach. But I think that since my answer to:

> Do we plan on introducing additional tags like, ever?

is: I really can't think of what we'd introduce, especially since we have `sc-custom`, and because we have `sc-custom` it would have to be a generic case, and therefore something we missed and something no one has reported an error on yet, as this will cover all issues related to this ever posted.

So I think I in good faith could release this, even with the danger of the semver-change.

I think another good thought on it is that releasing a new tag would never ever be urgent because we have `sc-custom`. So it could easily be queued with something else and put out later if there is something (such as `sc-soda` ;) ) that we realized, oh, that'd be really handy, though as I said I really can't think of what it would be.

Maybe everyone just go in the think tank once again, look through old css, and see if there are any other cases?

README.md
emilgoldsmith: Sure thing, I can update both of them

mxstbr:
Readme is cleaned up with the comments above.

----

I agree that `sc-custom` will solve most issue with edge cases, and to be honest major versions are just as cheap as minor or patch versions so it's not worth shipping a worse API for that reason.

This is good to go I think, great work!

emilgoldsmith:
> This is good to go I think, great work!

Yes I think I agree! Just did an extra quick double check. I'll just squash a few commits for a nicer history and then I'll merge it in :).

I'll also write a few things on Spectrum in the thread we had about going to beta regarding my thoughts on moving forward for the 1.0 release :)

ismay:
Tiny spelling error in the readme. Can I still push a commit to fix it or are you already squashing commits etc.?

emilgoldsmith:
> Tiny spelling error in the readme. Can I still push a commit to fix it or are you already squashing commits etc.?

Just finished now, you can add it now if you want and I can squash it again? Also feel free to take some more time to give the whole thing another review if you want @ismay

ismay:
> Just finished now, you can add it now if you want and I can squash it again? Also feel free to take some more time to give the whole thing another review if you want @ismay

Ok done. It's a supertiny fix, but anyway :). Other than that no remarks 👍

emilgoldsmith:
> Ok done. It's a supertiny fix, but anyway :). Other than that no remarks :+1:
Yep and I just squashed that one in as well :). So let's wait for the CI to pass and then merge it in :)

PR #74, opened 2017-07-19
[FIX] Handle interpolations in nested blocks

### LINKED ISSUE #73: Unknown word - CssSyntaxError
Hi!
Not sure if this is bug since I wasn't the one who set up our tooling.
But after upgrading to 0.2.0 yesterday I'm no longer able to use the css helper in nested rules.
Since I don't know how to properly set up a fresh environment I can't really prove it's a problem with the stylelint processor. But this was the only package I updated yesterday besides postcss.
 
```
const css = css`
  color: #FFF;
`

const div = styled(div)`
  span {
    ${css}
  } 
`
``` 

Feel free to close if you don't think it's a bug.

emilgoldsmith:
Had missed a case in #59, this is the breaking tests and the fix. This should also resolve #73, we can probably merge this in quickly and publish a 0.2.1 so our friend in #73 doesn't have to see ugly errors anymore :).

### DIFF
--- a/src/utils/general.js
+++ b/src/utils/general.js
@@ -49,6 +49,7 @@ const isLastDeclarationCompleted = text => {
   if (
     lastNonWhitespaceChar === ';' ||
     lastNonWhitespaceChar === '}' ||
+    lastNonWhitespaceChar === '{' ||
     lastNonWhitespaceChar === null
   ) {
     return true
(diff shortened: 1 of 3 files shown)


mxstbr:
Feel free to publish a new version! (you should have npm access)

emilgoldsmith:
Done! Thanks again for that, I really appreciate the trust @mxstbr :)

PR #53, opened 2017-06-29
WIP Failing test with incorrectly reported indentation errors

ravinggenius:
Failing test case for #51.

PR #77, opened 2017-08-01
Fix indented TTL case

### LINKED ISSUE #51: Expected indentation of 2 spaces
### Note

I originally opened this as an [issue for stylelint](https://github.com/stylelint/stylelint/issues/2646), but they directed me over here. I have reproduced the issue below. I am using `stylelint-processor-styled-components` v0.1.2.

### Issue

> Describe the issue. Is it a bug or a feature request (new rule, new option, etc.)?

I believe this is a bug. When linting [styled-components](https://www.styled-components.com/) styles inside a helper function, indentation is reported incorrectly.

> Which rule, if any, is this issue related to?

`indentation`

> What CSS is needed to reproduce this issue?

```javascript
import { css } from 'styled-components';

const helper = (condition) => {
  if (condition) {
    return css`
      color: red;

      &:hover {
        color: blue;
      }
    `;
  }
  return null;
};
```

> What stylelint configuration is needed to reproduce this issue?

```json
{
  "extends": "stylelint-config-standard",
  "syntax": "scss"
}
```

> Which version of stylelint are you using?

`7.11.1`

> How are you running stylelint: CLI, PostCSS plugin, Node API?

`yarn run lint` which calls `stylelint 'src/**/*.js?(x)'`

> Does your issue relate to non-standard syntax (e.g. SCSS, nesting, etc.)?

Yes. Both styled-components and SCSS look for `&` as a placeholder when expanding nested rules.

> What did you expect to happen?

No errors to be reported. An error only gets reported when both the `color: red;` style and the `&:hover` block is present. Removing either does not trigger an `indentation` error to be reported.

> What actually happened (e.g. what warnings or errors you are getting)?

The following errors were flagged:

```shell
Button.jsx
 36:5  ✖  Expected indentation of 2 spaces   indentation
 38:5  ✖  Expected indentation of 2 spaces   indentation
 39:7  ✖  Expected indentation of 4 spaces   indentation
 40:5  ✖  Expected indentation of 2 spaces   indentation
 41:3  ✖  Expected indentation of 0 spaces   indentation
```

emilgoldsmith:
This resolves #51.

Problem was that when replacing empty lines we were assuming base indentation was always 2 spaces, so in the min function it would always find 2 as the minimum if base indentation was actually more.

I changed it to assuming that the last backtick of the TTL is put on the level of the "base indentation level" which I think is a quite fair assumption, any better ideas?

Also will just put a comment in the code in one place when I submit the PR for something I was curious about.

I think it could probably be merged in now, could be I put in a few more tests tomorrow for oneline stuff etc. ? and more specifically for the issue that was found to be the root instead of just the specific case submitted in the issue.

### DIFF
--- a/src/utils/general.js
+++ b/src/utils/general.js
@@ -8,11 +8,17 @@ let count = 0
  */
 const fixIndentation = str => {
   // Get whitespaces
-  let match = str.match(/^[ \t]*(?=\S|$)/gm)
-  // Remove first, empty item
+  const match = str.match(/^[ \t]*(?=\S|$)/gm)
+
+  // Handle oneline TTL case
+  if (match.length === 1) {
+    return {
+      text: str,
+      indentColumns: 0
+    }
+  }
+  // Remove first, empty item, and if it is a oneline TTL we already handled it
   match.splice(0, 1)
-  // Remove empty lines
-  match = match.map(x => (x.length < match[match.length - 1].length ? '  ' : x))
 
   if (!match) {
     return {
@@ -21,14 +27,14 @@ const fixIndentation = str => {
     }
   }
 
-  // Get the minimum amount of indentation
-  const indent = Math.min(...match.map(x => x.length))
-  const re = new RegExp(`^[ \\t]{${indent}}`, 'gm')
+  // Get the base level of indentation assuming standard javascript TTL style
+  const baseIndentationLength = match[match.length - 1].length
+  const re = new RegExp(`^[ \\t]{${baseIndentationLength}}`, 'gm')
 
   return {
     // Remove the min indentation from every line
-    text: indent > 0 ? str.replace(re, '') : str,
-    indentColumns: indent
+    text: baseIndentationLength > 0 ? str.replace(re, '') : str,
+    indentColumns: baseIndentationLength
   }
 }
 
--- a/src/utils/general.js
+++ b/src/utils/general.js
@@ -10,30 +10,23 @@ const fixIndentation = str => {
   // Get whitespaces
   const match = str.match(/^[ \t]*(?=\S|$)/gm)
 
-  // Handle oneline TTL case
-  if (match.length === 1) {
-    return {
-      text: str,
-      indentColumns: 0
-    }
-  }
-  // Remove first, empty item, and if it is a oneline TTL we already handled it
-  match.splice(0, 1)
-
-  if (!match) {
+  // Handle oneline TTL case and empty line etc.
+  if (!match || match.length <= 1) {
     return {
       text: str,
       indentColumns: 0
     }
   }
 
-  // Get the base level of indentation assuming standard javascript TTL style
+  // We enforce that final backtick should be at base indentation level
   const baseIndentationLength = match[match.length - 1].length
-  const re = new RegExp(`^[ \\t]{${baseIndentationLength}}`, 'gm')
-
+  // Remove whitespace on empty lines before reindenting
+  const emptyLinesHandled = str.replace(/^[ \t]$/gm, '')
+  // Normalize indentation by removing common indent
+  const re = new RegExp(String.raw`^[ \t]{${baseIndentationLength}}`, 'gm')
+  const reIndentedString = emptyLinesHandled.replace(re, '')
   return {
-    // Remove the min indentation from every line
-    text: baseIndentationLength > 0 ? str.replace(re, '') : str,
+    text: reIndentedString,
     indentColumns: baseIndentationLength
   }
 }
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ npm run lint:css
 Turning off rules with `stylelint-disable`-like comments (see the [stylelint documentation](https://stylelint.io/user-guide/configuration/#turning-rules-off-from-within-your-css) for all allowed syntax) is fully supported inside and outside of the tagged template literals, do note though that what actually happens behind the scene is that all `stylelint-(disable|enable)` comments are moved into the compiled css that is actually linted, so something like this:
 
 
-```
+```js
 /* stylelint-disable */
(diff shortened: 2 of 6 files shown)


src/utils/general.js
```diff
@@ -8,11 +8,17 @@ let count = 0
...
+      text: str,
+      indentColumns: 0
+    }
+  }
+  // Remove first, empty item, and if it is a oneline TTL we already handled it
   match.splice(0, 1)
-  // Remove empty lines
-  match = match.map(x => (x.length < match[match.length - 1].length ? '  ' : x))
 
   if (!match) {
```
emilgoldsmith: What is this conditional here for? We already assumed above in `match.splice` and `match.length` that `match` exists, right? So shouldn't we either be checking this above as in, if there was no match at all, or not check it at all? The `match.splice(0, 1)` won't ever make the array `undefined` or `null` will it? Not from my experience or from a quick look at documentation anyway.

ravinggenius:
> could be I put in a few more tests tomorrow for oneline stuff

While I was reducing the failing helper function, I noticed it wasn't breaking unless I added an empty line in the styles:

```javascript
if (condition) {
  return css`
    color: red;
  `
}
```

... worked fine. So did:

```javascript
if (condition) {
  return css`
    &:hover {
      color: blue;
    }
  `
}
```

Thanks for getting into this!

src/utils/general.js
mxstbr: This should probably be `match.length === 0`? Even empty arrays are truthy so this condition is never hit.

emilgoldsmith:
Yeah @mxstbr, I did think of a single a bit weird case of an unfinished declaration block like an interpolation that does something like this:
```js
const bla = css`
  color: blue;
  &:hover {
    color: red;`
```
But it seems very very edge casey indeed, and if we actually have people filing issues about something like this I guess we can always add a note to the readme, but I feel like this would be so weird to do that yeah, I'd be surprised if we ran into problems with it.

src/utils/general.js
emilgoldsmith: Yeah exactly, I think I'll fix that up, and maybe also just add back a bit of stuff to handle empty lines, and a bit more thorough tests.

emilgoldsmith:
>  I noticed it wasn't breaking unless I added an empty line in the styles:
@ravinggenius I'm not quite sure what you mean by that exactly? Is it something I should be aware of and take into account in tests / fix?

ravinggenius:
@emilgoldsmith I was just noting that my personal issue wasn't triggered unless the blank line was in the styles. In case you were trying to reduce the test case, I think I already did that.

emilgoldsmith:
@ravinggenius okay, thanks for trying to help :). I'm still not quite sure what you mean by `the blank line` though. If you want to elaborate a bit feel welcome, but otherwise I'm also pretty sure I've reduced it, so no problem. But more information is always helpful if you have any I may not have caught :).

emilgoldsmith:
Just made it WIP because I've been playing a bit in my spare time with some different ideas that are more clean / handle more edge cases, think I'm nearly there but now I actually implemented a solution that autofixes too much so it doesn't catch some indentation errors that maybe should be caught? I'll come back to it again a bit later and maybe put up an example here for discussion.

emilgoldsmith:
@mxstbr @ismay So the way I see it we may have to make a few decisions based on the following two examples:
Case 1:
```js
const button = styled.button`
      color: red;`
```
Case 2:
```js
if () {
  if () {
    const button = styled.button`
      color: red;`
  }
}
```

We can't distinguish between the above two examples as they are completely equal in terms of the actual template literal, but at least in terms of our current tests + a bit of common sense we would currently say the top one is wrong indentation and the bottom one correct according to normal indentation rules.

It would also be the same problem if both of them put the closing backtick on the line below but the bottom one also puts it on column 1 (as in unindented).

The way I see it we 100% should support case 2 being correct, which is also the problem @ravinggenius  brought up in his issue. So from my point of view this leaves us with two options:
1. I currently have an implementation in place (not pushed) that normalizes all the indentation perfectly, but it also normalizes Case 1 above which currently breaks one of our tests in an example of bad indentation.
2. We enforce our own custom "stylelint TTL style" which we would of course put a note about in the README that you should always put the closing backtick on the following line, and put it in the level of your current indentation level.

I guess as a future possible customization thing for option 2 you could maybe add an option somehow that would decide which style you wanted but I don't think that's even close to high priority, especially as if we go with option 2 I think most people probably already do the that, and that would be the style to enforce.

What are your thoughts?

mxstbr:
> We enforce our own custom "stylelint TTL style" which we would of course put a note about in the README that you should always put the closing backtick on the following line, and put it in the level of your current indentation level.

I like this, I vote for that 👆

ravinggenius:
@emilgoldsmith I was referring to my original issue/test case when I mentioned the "blank line". See https://github.com/styled-components/stylelint-processor-styled-components/pull/53/commits/f5fa473bdf9efcd4c633244a08fd4f5eb747c7ac#diff-262b487a27f134c9a6af631adc904cd6R42. I should have been more clear.

emilgoldsmith:
Okay this is done now IMO, updated the README and finished + simplified implementation of `fixIndentation`. If @styled-components/stylelint-processor wants to take a second look at it as it changed a bit though I don't imagine there should be anything shocking in there, and we can put it live.

I believe this should probably be in 0.3 as the change as described in the README is a breaking change.

emilgoldsmith:
Awesome, let's merge it :). I just rebased onto master so when checks are done I'll merge it in.

emilgoldsmith:
100% coverage let's go ;) (the only thing lacking was apparently removing the weird if statement we talked about above that never runs)

mxstbr:
NICE!!

PR #98, opened 2017-08-15
Add reference to stylelint-custom-processor-loader to README

emilgoldsmith:
I think my Webpack loader is ready to see the light with core features etc. at least. I'd love to add it to the README here in a combination of shameless promotion of my own npm module and also just because I know I really wanted this when I started using our processor.

P.S: I stole your good idea of the gif in the README but upgraded it with the hot reloading webpack allows ;).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,8 +59,13 @@ Now you can lint your CSS by running this script! 🎉
 npm run lint:css
 ```
 
+### Webpack
+
+For use with Webpack you can use the [`stylelint-custom-processor-loader`](https://github.com/emilgoldsmith/stylelint-custom-processor-loader).
+
 ### Syntax notes
 #### Turning rules off from within your CSS
+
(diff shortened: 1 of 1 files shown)


emilgoldsmith:
> It looks great!

Thanks!

>  do you want to move that webpack loader over to this organisation?

I think I'll keep it separate for now, for one because while it was definitely written for Styled Components, I did see a void in general for custom processors of Stylelint, and it could be that there are some use cases for some of the other custom processors, or maybe new cool things like Styled Components will be invented in the future that might want to use it :).

And then I'm also a bit proud and protective of it as it's my first npm module so I think my pride wants to keep it on my page for now :P. 
If at some point in the future it for some reason would make a lot of sense to migrate it I'd be open to it, but for now I also see reasons like the above to keep it separate :).

mxstbr:
That make sense!

PR #101, opened 2017-08-17
Move typescript to devDependencies

### LINKED ISSUE #99: Remove typescript from dependencies
A place to continue the discussion from https://github.com/styled-components/stylelint-processor-styled-components/pull/49#issuecomment-323036638:

> Still feels weird to bundle typescript as a dependency btw, when it seems more appropriate as a peerDependency. Maybe good to discuss that in another issue, don't know how you feel about that?

> Yeah why is it it's a dependency by the way? It shouldn't need to be a peerDependency either should it? As in, if you don't use Typescript you shouldn't need it, couldn't we just make it a devDependency since the only reason we have it here is for testing, right? Or am I missing something?

> I think we did that because typescript-eslint-parser throws if typescript can't be found (or complained anyway). Which is annoying for people who don't use our processor for typescript as they have no need for typescript in their project. So we're basically stuck with this situation unless it can somehow be fixed on typescript-eslint-parser's end (by not doing the check until it's actually instantiated or something, don't know what they're doing exactly).

> Hmm that's a pain <.<. When I'm done reviewing this which should be in a sec I'll just look at our code regarding Typescript as I have never looked at that, mostly out of curiosity, would be awesome if we could remove that dependency somehow for sure though it doesn't sound like it.

emilgoldsmith:
I found a solution :). As long as you guys are okay with conditional requires. This resolves #99

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5324,9 +5324,10 @@
       "dev": true
     },
     "typescript": {
-      "version": "2.3.4",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-2.3.4.tgz",
-      "integrity": "sha1-PTgyGCgjHkNPKHUUlZw3qCtin0I="
+      "version": "2.4.2",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-2.4.2.tgz",
+      "integrity": "sha1-+DlfhdRZJ2BnyYiqQYN6j4KHCEQ=",
+      "dev": true
     },
     "typescript-eslint-parser": {
       "version": "5.0.0",
--- a/package.json
+++ b/package.json
@@ -47,12 +47,12 @@
     "lint-staged": "^4.0.1",
     "lodash": "^4.15.0",
     "prettier": "^1.4.4",
+    "typescript": "~2.4.0",
     "stylelint": "^8.0.0"
   },
   "dependencies": {
     "babel-traverse": "^6.16.0",
     "babylon": "^6.12.0",
-    "typescript": "~2.3.2",
     "typescript-eslint-parser": "^5.0.0"
   }
 }
--- a/src/parsers/index.js
+++ b/src/parsers/index.js
@@ -1,5 +1,4 @@
 const estreeParse = require('./babylon-parser')
-const typescriptParse = require('./typescript-parser')
 
 const traverse = require('babel-traverse').default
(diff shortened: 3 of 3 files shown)


ismay:
Ah nice. And the typescript peerDependency requirement from typescript-eslint-parser still gets logged when installing this processor right? Because that also makes it clear to the end-user that we require a specific version of typescript, should they want to use it.

emilgoldsmith:
> Ah nice. And the typescript peerDependency requirement from typescript-eslint-parser still gets logged when installing this processor right?

I tried uninstalling typescript completely (just for tests) and ran the non-typescript tests and it didn't work without the dynamic require, did with it, and typescript tests work with dynamic require as well when typescript is then installed of course :). It still logs the annoying "you are not using the intended version" thing when you use it, I'm not sure about on install though. I can see that in a `package.json` since they actually [only put any version of Typescript](https://github.com/eslint/typescript-eslint-parser/blob/ab4e05e0e90cb8a5a21a92464beca9e779a2f036/package.json#L60) as their peerDependency. So they enforce the version only in those annoying logs.

ismay:
>  I can see that in a package.json since they actually only put any version of Typescript as their peerDependency. So they enforce the version only in those annoying logs.

Hmm, yeah I guess that is so they can leave the final decision to the user, if they want to try a different version it might still work.

emilgoldsmith:
> Hmm, yeah I guess that is so they can leave the final decision to the user, if they want to try a different version it might still work.

Yeah, I'm not quite sure what current behaviour would be during installation to be honest, I just tried doing an install now of the package now that didn't give any warning... so I'm not quite sure. But given the behaviour during the tests with the warning logs coming up about recommended version I'm pretty sure that should happen when you use our package as well. 

I think I'm personally okay with this to the extent that I'd merge this in and think it shouldn't cause any problems. If it does I guess we'd hopefully get some issues submitted so we could become aware of it, but I think this should be okay.

ismay:
> I think I'm personally okay with this to the extent that I'd merge this in and think it shouldn't cause any problems. If it does I guess we'd hopefully get some issues submitted so we could become aware of it, but I think this should be okay.

Yeah I agree 👍

emilgoldsmith:
Okay, let's merge it in then, this is pretty minor, mostly for our own preferences so I'll just put it in master and it can go with some future publish, maybe even 1.0 if we don't publish anything else before then.

PR #105, opened 2017-08-22
Greenkeeper/typescript eslint parser 7.0.0

emilgoldsmith:
Closes #103 and also closes #102


emilgoldsmith:
I also checked that the recommended version of TS didn't change so as to update that if so so we don't get the annoying logs when running tests, and it's all good so I think I'll just merge this in.

PR #107, opened 2017-08-26
Support Styled Component properties

### LINKED ISSUE #106: Not reporting errors when adding attaching additional props
I would expect this to fail:

```js
const StyledLightBulbIcon = styled(Icon).attrs({ iconName: 'light_bulb' })`
  svg-bad-selector {
    height: 50px;
  }
`
```

However it doesn't.  I passes without an issue.  When I remove the call to the `.attrs` function, it fails with an error.

```js
const StyledLightBulbIcon = styled(Icon)`
  svg-bad-selector {
    height: 50px;
  }
`
```

```
src/components/task/list.jsx
 10:3  ✖  Unexpected unknown type selector "svg-bad-selector"   selector-type-no-unknown
```

Is there an option that can be passed into the CLI to ensure that it validates the styles when the `.attrs` function is being used?

emilgoldsmith:
I have currently implemented a fix for #106 (`.attrs`) but after double checking the Styled Components API I also want to put in a similar check for `.extend` before merging it in.

EDIT:
Done, this should add support for both `.extend` and `.attrs`, and is ready for review now.

resolves #106

### DIFF
--- a/src/parsers/index.js
+++ b/src/parsers/index.js
@@ -6,6 +6,7 @@ const isHelper = require('../utils/styled').isHelper
 const isStyledImport = require('../utils/styled').isStyledImport
 const hasAttrsCall = require('../utils/styled').hasAttrsCall
 const getAttrsObject = require('../utils/styled').getAttrsObject
+const isExtendCall = require('../utils/styled').isExtendCall
 
 const wrapSelector = require('../utils/general').wrapSelector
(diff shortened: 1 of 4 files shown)


emilgoldsmith:
Ready for review

emilgoldsmith:
I'm thinking I'll just publish a 0.4 with this quickly so our friends can benefit from this bug fix? :)

mxstbr:
Sure!

PR #111, opened 2017-08-28
Merge 1.0 branch into master

mxstbr:
**THIS DOES NOT RELEASE v1.0**

Since I don't think we'll do anymore pre-v1 releases we can now merge the `1.0` branch into master to avoid them going out of date all the time. WDYT?


mxstbr:
Ugh linting... One sec.

mxstbr:
Now prettier fails, even though it doesn't locally. That's fun!

ismay:
> Now prettier fails, even though it doesn't locally. That's fun!

Maybe version difference between your local version and the remote one?

mxstbr:
I thought of that, but I'm running the latest version :confused:

It's still failing and now I'm out of ideas why that could be. Ugh.

ismay:
@mxstbr This should fix it. Running `lint:prettier` showed the same error as on travis for me locally. Should be a version mismatch or just a file that wasn't fixed by prettier previously. Anyway, hopefully it passes now.

emilgoldsmith:
I think it could be related to node versions possibly?

mxstbr:
Thanks @ismay!!

PR #112, opened 2017-08-28
Add moduleName option

### LINKED ISSUE #109: Make import name configurable
Right now this processor is hard-coded to only work with `styled-components` due to checking the import name here:

https://github.com/styled-components/stylelint-processor-styled-components/blob/f67b19276f6fc252af244d60d3e69fe8e8159a3b/src/utils/styled.js#L6-L7

We should add an option (`importName`? `library`? `package`?) to allow users to say that the name of the imported package is different so it can be used with `styled-jss`, `emotion`, etc.

It'd be a shame if all those libraries had to build their own processor, when really they only have to change that one string and it'd work. I don't think we should promise any particular level of support, since all those libraries have slightly diverging APIs, but having basic support would be prettttttty great. (and also awesome to get out with v1.0)

mxstbr:
**MERGE #111 FIRST**

Closes #109


src/utils/styled.js
```diff
@@ -3,8 +3,8 @@ const isTaggedTemplateLiteral = require('./tagged-template-literal').isTaggedTem
 /**
  * Check if something is a styled-components import declaration
  */
-const isStyledImport = node =>
-  node.type === 'ImportDeclaration' && node.source.value === 'styled-components'
+const isStyledImport = (node, importName) =>
+  node.type === 'ImportDeclaration' && node.source.value === importName
```
emilgoldsmith: Hmm in the README you're saying this depends on the name you import it with, but here we check the library imported, so you should probably update that in the README. Our code is actually agnostic to the variable names you assign them to, as far as I understand anyway. So therefore if you want to use it with another library you would have to specify that option, as we never check for variables called `styled` I believe.

In that case the argument here and the option name should probably be `libraryName`, yeah?

mxstbr: Read the associated issue, I just phrased the README badly.

We default to `styled` for the assigned variable name of the default export, so it'll work if you do `import styled from 'emotion'`, and we also check the imports so it works if you do `import other from 'styled-components'`, but at the moment if you do `import other from 'emotion'` it'll break. That's what this PR fixes, basically.

mxstbr: I remembered why I called it `importName` added my reasoning in the above comment. I can see why `importName` could be confusing, since "name" doesn't unambigously point to either parts of the import statement.

Babel calls it `importNode.source.value`, so maybe `importSourceName`?

mxstbr: Maybe I'm thinking this through too much :thinking:

emilgoldsmith: Hmm okay, I think I get you now. My main points of misunderstanding was mostly from this line in the README:
> If you want to avoid edge cases when you assing the default import to a variable not called "styled"

As I think that's super misleading as it actually has nothing to do with what you name your the variable you import it into (it would also work, as I know now you didn't have an understanding problem with, if you did `import wabalubadubdub from 'styled-components`). 

And then I also had a problem because I never heard of the webpack aliasing thing because I never used it, when you talked about aliasing I thought you might've meant `import { injectGlobal as some_alias }` or stuff like that which again wouldn't have been relevant but I researched the webpack alias thing, and get you now!

emilgoldsmith:
Okay, README updated with a suggestion for change :)

mxstbr:
I like the change, it's not entirely accurate though, because all these cases work at the moment:

```JS

// ✅ This is linted! 
const Button = styled.div`
  color: #PP0;
`

// ----

import somethingElse from 'styled-components';

// ✅ This is linted! 
const Button = somethingElse.div`
  color: #PP0;
`

// ----

import styled from 'some-other-lib';

// ✅ This is also linted! 
const Button = styled.div`
  color: #PP0;
`

```

It's just that the case breaks where you change the import name of a different library:

```JS
import somethingElse from 'some-other-lib';

// ❌ This isn't linted
const Button = somethingElse.div`
  color: #PP0;
`
```

So that's what I'm trying to fix here.

----

Re: naming, I agree with your point, I think "path" makes it a bit confusing though. What do you think about `"moduleName"`?

emilgoldsmith:
Interesting, because my whole "disagreement" with some of the things you wrote here is based on me being pretty sure that we don't lint the last case you gave, this one:

```
import styled from 'some-other-lib';

// ✅ This is also linted! 
const Button = styled.div`
  color: #PP0;
`
```

Are you really sure we do that? Can you point me to the line of code that picks up on this? Because `isStyledImport` just checks if the imported name is `styled-components` so that's what I've based my thoughts on in this discussion.

mxstbr:
I'm fairly certain the `isStyledImport` thing only checks it to change the variable name that's imported but doesn't break if it doesn't find the library? Maybe that's incorrect though?

emilgoldsmith:
I'm like 95% sure you're incorrect. That's also why I've been correcting you this whole time, try and test it, I'm really sure it doesn't care about absolutely anything but the `importPath` or w/e we call it, as we use the `importNames` or something like that dictionary that keeps a map of what the variable names actually are.

I could be wrong but I'd be super surprised from my current understanding of our codebase

mxstbr:
Okay, I'm adding a test right now!

mxstbr:
I've added a test in 483ee49 and it passes. (at least locally) Did I do something wrong with it?

mxstbr:
I'm pretty certain that's the way it works, honestly. I'd be surprised to find out it doesn't work that way. The only thing the `isStyledImport` thing does is check for the `import other from 'styled-components'` case, not for the actual linting.

emilgoldsmith:
Okay I get it now, you are right, but yeah, it's kinda subtle. Since we have default values for the `importedNames` object and we never overwrite anything due to `isStyledImport(node)` never evaluating to true the defaults are kept, and the default is, as you say, `styled` so I now understand that it is indeed how it works and why.

On a different note though (it's a different discussion whether this is high priority) I don't think I like that implementation though, so right now we don't care about anything but the variable name? So if anyone calls any TTL with the names `styled`, `css`, `keyframes`, or `injectGlobal` we will be linting whatever is inside? I find that very bad behaviour. In this case even
```js
const styled = (text) => console.log(text)
styled`hello, what's up`
```
would be linted and throw a lot of errors. IMO we should not have defaults and only instantiate `importedNames` if `isStyledImport` returns true and gives us the actual imported variable names, and then also implement the option we talk about in this PR, and then in that case my README would also be correct as this is how I thought it worked before.
But that's just my 2 cents of course.

mxstbr:
I decided to go down this route because I can't prescribe what/if people use module bundlers. Some folks might use global variables, and that's perfectly fine!

The reason I kinda doubt that this is an issue is 1) because it's never come up and 2) if you put code like this in your codebase that's linted with the styled components processor maybe you deserve an error:

```JS
styled.div`
  not css here this is just a string
`

css`
  nah mate, css is not here
`

injectGlobal`
  yeah no this is definitely not CSS
`
```

The only one I'm 50/50 about is `injectGlobal` where I could _maybe_ see people using that name for something else but it's so highly unlikely that I think it's worth not being finnicky about build systems and module bundlers.

Happy to talk about it if you disagree though, let's open an issue?

----

To get back on track, what name do we go with now? :laughing:

emilgoldsmith:
Yeah okay, I guess I can see some logic in this. Maybe I'll add an explanatory comment to the code the next time I'm touching `parsers/index.js` as I think the intention behind it could be a bit clearer in the code :). Thanks for taking the time for this long discussion haha.

----

For the name.... yes... `moduleName` might be quite good actually. I dunno, it doesn't quite take into account the path stuff, but yeah, you know we could even do a small implementation thing where we also check if the last part of the path equals the module name so one didn't need to do something with a path if you for some reason used a path. Like a check like `ORIGINAL CHECK || source.split('/')[source.split('/').length-1] === moduleName` kinda thing (and do another || and check with \ as well). That would both be nicer behavior in terms of user experience I feel, and also it could fit well with naming it `moduleName` as it could then be path ambiguous.
What do you say to that?

----

And for another one! I think I'll just write about this in Spectrum, but I feel like we're pretty much ready to publish 1.0 now right? We can continue this in Spectrum

mxstbr:
I forgot about relative paths, but I like that. I'll go ahead and implement this quick!

mxstbr:
Oh wait, gotta update the docs.

mxstbr:
Submitted PR to add the documentation for this option to the website here: https://github.com/styled-components/styled-components-website/pull/107

src/utils/styled.js
```diff
@@ -3,8 +3,9 @@ const isTaggedTemplateLiteral = require('./tagged-template-literal').isTaggedTem
 /**
  * Check if something is a styled-components import declaration
  */
-const isStyledImport = node =>
-  node.type === 'ImportDeclaration' && node.source.value === 'styled-components'
+const isStyledImport = (node, moduleName) =>
+  node.type === 'ImportDeclaration' &&
+  node.source.value.split('/')[node.source.value.split('/').length - 1] === moduleName
```
emilgoldsmith: Yeah this is good :).

The only thing is maybe we should also do an `||` with `\` for Windows? Or maybe the `path` module can do something like this for us where it handles all the edge cases?

mxstbr:
Awesome!

PR #114, opened 2017-08-30
Update Readme with links to website

SaraVieira:
As discussed here:
https://github.com/styled-components/styled-components-website/issues/78


ismay:
@mxstbr @emilgoldsmith I understand the need for discoverability and therefore putting the docs on the styled components website, but one thing I am afraid of is that now the docs are no longer coupled to the code. So checking out a certain commit would not include documentation (i.e. a readme.md) related to the code in that commit. As well as delays in updating the docs since it now lives in two different repos.

Interested in what your thoughts on that are. Other than that this pr looks good.

mxstbr:
I've thought about that, back when we moved the Keystone docs it was a bit of a pain to keep the two repos in sync so I'm well aware of the issue!

In this specific case I just think the docs aren't going to change super much; it's not like how to use the processor will fundamentally change. In my opinion discoverability is more important than that, not enough people use the processor and the big reason is few know it exists. If you're not happy with that I'm happy to go back to the other setup, from my PoV this makes a lot more sense!

ismay:
> In this specific case I just think the docs aren't going to change super much; it's not like how to use the processor will fundamentally change.

Yeah true.

> In my opinion discoverability is more important than that, not enough people use the processor and the big reason is few know it exists.

Ok, sounds like a good tradeoff then. And it's just a minor thing. Thought I'd just ask what the opinions were on this. Thanks for elaborating 👍

emilgoldsmith:
Yeah, I'm post-flight back online and while this is already merged I agree with the argument that post publishing v1.0 the README should have pretty few changes, but otherwise @ismay's argument was also very relevant.

PR #116, opened 2017-08-31
Remove selector wrap from injectGlobal

### LINKED ISSUE #26: Don't prefix injectGlobal with a selector
This code...
```js
injectGlobal`
  html {
    color: blue;
  }
`;
```

... with `selector-max-compound-selectors` rule set to `1` returns this validation issue
```sh
Expected ".selector html" to have no more than 1 compound selectors
```

emilgoldsmith:
resolves #26.

Put this together on the plane, could possibly use a few more tests but I think it should be pretty okay?


test/fixtures/simple/global.js
```diff
@@ -12,13 +12,6 @@ const animation = keyframes`
 }
 `
 
-// ⚠️ Wrong indentation ⚠️
-injectGlobal`
-html {
-  margin: 0;
-}
-`
-
```
mxstbr: That should still be an error, that's wrong indentation. Just like we don't support it with `css` or `styled`.

emilgoldsmith: Oh yeah, I forgot to comment on this as I wrote the code yesterday. The problem is my `removeBaseIndentation` function and I don't see any good way to go around this, do you understand what I mean or should I be more explicit?

mxstbr: Not quite, why shouldn't this throw an error anymore?

emilgoldsmith: Yeah sorry I'm also a bit jetlagged and sitting in an 8 hour orientation day at NYU but I'll do my best :P.

In order to avoid general indentation errors I need to remove all the base indentation, right? So as long as that function is being called this won't throw an error. And in order to make it throw an error we would somehow need to induce an error as if they don't indent anything like that we actually just want to straight up insert it. It's purely a Styled Components style problem, not a css problem.

I'm not saying we shouldn't, I'm saying it's maybe too hard

ismay: > In order to avoid general indentation errors I need to remove all the base indentation, right?

Is it not the goal here to have exactly the same output as what styled-components' injectGlobal would output? Because that's basically what we want to lint right?

emilgoldsmith: Yep, that's a really good point to make @ismay!

I'm not myself sure exactly with all details what styled components would output though, we could probably even just replicate the code used in styled-components actually

mxstbr: So if you do:

```JS
injectGlobal`
  html {
    color: blue;
  }
`
```

styled-components injects

```CSS
html {
  color: blue;
}
```

(duh)

We should lint this as an indentation error is what I'm trying to get at:

```JS
injectGlobal`
html {
  color: blue;
}
`
```

That's wrongly indented and thusly should throw an indentation error if that rule is enabled.

emilgoldsmith: Yeah I understand that, but how would we fix it? Are we going to somehow artificially inject an error for this special case? And also would `styled-components` inject wrong CSS if you wrote it like that?

ismay: > And also would styled-components inject wrong CSS if you wrote it like that?

Exactly, that's what I'm wondering as well. Can't test it because I'm on mobile, but would it not insert the same css as the properly indented one?

mxstbr: It would insert the same CSS, just like this code would:

```JS
const Button = styled.div`
color:  blue;
`
```

Is that correctly indented? No, so the indentation rule in stylelint kicks in and warns the user that it's wrongly indented.

The same thing should happen with `injectGlobal`, at least I thought so. Why shouldn't that be the case?!

ismay: > It would insert the same CSS

If both this code:

```javascript
injectGlobal`
  html {
    color: blue;
  }
`
```
and this code:

```javascript
injectGlobal`
html {
  color: blue;
}
`
```

will both inject this bit of css:

```css
html {
  color: blue;
}
```

Then that's both ok as far as stylelint is concerned right (if you have base indent at 0 and within block at 2 spaces)? The resulting css is identical. I see that one style looks ugly in js, but that gets lost in translation as far as the css is concerned. Maybe the ttl base indentation is more of an eslint concern.

But maybe I'm missing your point.

mxstbr: The whole point is that it looks ugly in JS! We aren't linting the resulting CSS we're linting the code the user wrote and we make sure it doesn't look ugly and has errors.

mxstbr: That's why

```JS
const Button = styled.button`
color: blue;
`
```

throws an error, eventhough the resulting CSS is the same it's still ugly and it's a stylelint error. That's the whole point of stylelint?

emilgoldsmith: Yeah and I get that part, and while @ismay might be making a slightly different point than mine, my reason for removing that test was simply that my only way of implementing this in a simple way broke that test, so I deleted it thinking of it as collateral damage of this PR, not because I wanted to delete it but because I "had to" (not really I guess, I haven't thought it out yet but it seems like it'd get really ugly writing that code to handle this special case, and the code could possibly be error prone as well and yeah...).

The thing is, in order to prevent all the other cases of the indentation rule breaking we need to put everything on the baseline anyway, so we'd have to somehow implement the stylelint indentation rule ourselves in order to check if before we put it on the baseline whether it has broken our semantics and then somehow put in a purposefully wrongly indented piece of css into stylelint to force it to throw an error or something like that and that get's messy reeaaaalll fast I'm afraid.

That's my side of the story.

mxstbr: Ahhh okay, that makes sense. In that case, let's log that somewhere and move on!

emilgoldsmith: Okay haha :)

emilgoldsmith:
@mxstbr so if we're at least temporarily okay with deleting the broken test I deleted is there anything else you'd like changed specifically for you to approve the PR and we merge it in?

mxstbr:
Just some tests that use `injectGlobal`, that's all!

emilgoldsmith:
> Just some tests that use injectGlobal, that's all!

Hmmm... maybe I could add one to `test/fixtures/simple/valid.js`, but something that isn't visible from the PR as that was already in place is that there are already a few fixtures with `injectGlobal` I just deleted one of them that broke but the rest are still there:
https://github.com/styled-components/stylelint-processor-styled-components/blob/bddac5bdd15568d289fe33affbc95836d3eb3e54/test/fixtures/simple/helpers.js
and
https://github.com/styled-components/stylelint-processor-styled-components/blob/bddac5bdd15568d289fe33affbc95836d3eb3e54/test/fixtures/simple/imports.js

it's not like that's super hardcore tested, but it is in there, and I adjusted the expect x amount of warnings to our new expected behavior, so apart from maybe adding an `injectGlobal` to `valid.js` I'm not quite sure what else I'd add, does that sound okay?

mxstbr:
Just add the code from the original issue that we're trying to fix as a test:

```JS
injectGlobal`
  html {
    color: blue;
  }
`;
```

to make sure we don't regress :blush: That's all!

emilgoldsmith:
Okay, I'll get it over and done with right away :)

emilgoldsmith:
Okay, I believe this should be ready now :).

There was more prettier weird errors like @mxstbr experienced recently and I also did in the start of my time contributing here, I had to do an `npm install` to be able to see the error as well, I'm not sure if I had an outdated version of prettier? Did we update it recently? Anyway I added the `package-lock.json` changes my npm install introduced and fixed the error that it seems was in code I hadn't even changed, maybe my old version of prettier had changed something that the new one didn't agree on or something.

Anyway, stuff should be good now.

I'll just rebase it now to get rid of the merge conflicts.

mxstbr:
-0.4% coverage :cry:

mxstbr:
This part in `removeBaseIndentation` isn't covered:

screen shot 2017-09-02 at 6 20 04 pm

Any ideas if this code is necessary/if we can test it?

emilgoldsmith:
Oh yeah, I'll just pop in a unit test covering that :)

mxstbr:
Awesome!! Let's open an issue for that indentation thing?

emilgoldsmith:
Wuhuu! Doesn't this actually mean we're ready for publishing 1.0 now? :D.

We'd of course update the changelog as part of the publish, but I actually think this is all that I personally wanted fixed before 1.0 anyway.

PR #119, opened 2017-09-02
Update README

### LINKED ISSUE #12: Typescript support
I don't know how feasible it is but thought that I should put it out here.
It would be awesome with Typescript support.

mxstbr:
Adds better setup instructions, adds note about TypeScript support.

Closes #12


mxstbr:
No it's because we set `flow: true` in the `babylon` options! :blush:

PR #132, opened 2017-10-05
Update Greenkeeper Ignore

emilgoldsmith:
Updated Greenkeeper Ignore to not update any of our devDependencies that aren't directly related to the performance of the repo. (also removed Typescript from the ignore as I guess it could potentially be important to track the changes of that for our Typescript support, yes?)

This would also close #120

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,14 @@
     "ignore": [
       "eslint",
       "eslint-plugin-import",
-      "typescript"
+      "eslint-config-airbnb-base",
+      "babel-cli",
+      "babel-preset-latest",
+      "husky",
+      "jest",
+      "lint-staged",
+      "lodash",
+      "prettier"
     ]
   },
   "devDependencies": {


ismay:
Btw., originally typescript was added to the ignores because otherwise greenkeeper might want to update typescript beyond what's supported by typescript-eslint-parser. Since according to their readme anything beyond `~2.5.1` isn't officially supported and rated as experimental.

Currently typescript is at 2.5.3, so not a big difference, but that was the rationale for adding it to the ignores, since we can't always automatically update to the latest version.

emilgoldsmith:
Ahh that makes sense, so would you like to put it back, or should we just be aware of that?

ismay:
I think it’s fine. Just good to keep an eye on it as it could cause tests to fail.

emilgoldsmith:
Okay, sounds good!

PR #139, opened 2017-11-04
Add .babelrc

### LINKED ISSUE #136: Error with Node v4
Hi,

I'm facing a problem running the linter with node 4, node 6 and 8 is working fine.

```bash
> $ yarn lint:css                                                                                                                                            ⬡ 4.8.4 [±master ✓]
yarn run v1.2.1
$ stylelint './src/**/*.tsx'
SyntaxError: Unexpected token {
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at module.exports (/Users/plitex/dev/react/app/node_modules/stylelint/lib/dynamicRequire.js:5:10)
    at /Users/plitex/dev/react/app/node_modules/stylelint/lib/augmentConfig.js:342:23
    at Array.forEach (native)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Stylelint config:

```json
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-styled-components"
  ],
  "syntax": "scss"
}
```

chinesedfan:
Fixes #136.

Without `.babelrc`, the current build script can't transform any codes at all. Node 4 doesn't support those new ES6 syntaxes, so errors were thrown.

### DIFF
--- a/test/simple.test.js
+++ b/test/simple.test.js
@@ -1,7 +1,7 @@
 const stylelint = require('stylelint')
 const path = require('path')
 
-const processor = path.join(__dirname, '../src/index.js')
+const processor = path.join(__dirname, '../lib/index.js')
 const rules = {
   'block-no-empty': true,
   indentation: 2,
--- a/test/simple.test.js
+++ b/test/simple.test.js
@@ -118,7 +118,7 @@ describe('simple', () => {
 
     it('should be indentation and "empty line before" warnings', () => {
       const warnings = data.results[0].warnings.reduce(
-        (all, { rule }) => (all.includes(rule) ? all : all.concat(rule)),
+        (all, { rule }) => (all.indexOf(rule) >= 0 ? all : all.concat(rule)),
         []
       )
 
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["latest"]
+}
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,9 @@
 {
-  "presets": ["latest"]
+  "presets": [
+    ["env", {
+      "targets": {
+        "node": 4
+      }
+    }]
+  ]
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - "4"
   - "6"
   - "7"
   - "8"
--- a/package-lock.json
+++ b/package-lock.json
@@ -347,9 +347,9 @@
       }
     },
     "babel-helper-define-map": {
-      "version": "6.24.1",
-      "resolved": "https://registry.npmjs.org/babel-helper-define-map/-/babel-helper-define-map-6.24.1.tgz",
-      "integrity": "sha1-epdH8ljYlH0y1RX2qhx70CIEoIA=",
+      "version": "6.26.0",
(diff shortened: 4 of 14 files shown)


emilgoldsmith:
Thanks for the PR @chinesedfan :). For one I think "latest" is deprecated though and been replaced by "env" right? And also don't you need to install the presets with npm? Like following [this guide](https://babeljs.io/docs/plugins/preset-env/)? Have you checked that it currently works?

chinesedfan:
@emilgoldsmith `babel-preset-latest` is already in `devDependencies`, so I set it without modifications to `package.json`. If you prefer `env` now, please tell me targets that need to be supported.

I notice only Node 6/7/8 are tested in Travis CI. Shall I really add Node 4? If not, then #136 should be ignored. But `.babelrc` is still needed.

emilgoldsmith:
@chinesedfan ahh sorry! I hadn't noticed we already had the preset installed :). And I'm guessing our tests probably wouldn't run on Node 4 since that runs on precompiled code right? So it's probably not worth it, unless you want to upgrade our CI so that it compiles it and runs our tests on the compiled code instead of the precompiled one (which we probably already should be doing to be honest).

@mxstbr @ismay any thoughts on this PR?

ismay:
If we want to support node 4 I’d add preset-env and transpile for node 4. It’s supported until april 2018, so we could add it.

chinesedfan:
@ismay @emilgoldsmith So conclusions are:

- Add Node 4 into CI
- Test based on compiled codes
- Compile by babel-preset-env

If no problems, I will update as above.

emilgoldsmith:
Yeah if you're up for that go for it!

chinesedfan:
@emilgoldsmith Updated. Please check when you are free.

test/simple.test.js
```diff
@@ -118,7 +118,7 @@ describe('simple', () => {
 
     it('should be indentation and "empty line before" warnings', () => {
       const warnings = data.results[0].warnings.reduce(
-        (all, { rule }) => (all.includes(rule) ? all : all.concat(rule)),
```
emilgoldsmith: If we're compiling everything correctly now, can't we use this `includes` ESNext syntax?

chinesedfan: @emilgoldsmith No. Babel only transform codes in `src`, but not `test`.

chinesedfan:
@emilgoldsmith I updated the test script and didn't know why the CI failed at the first time. So I triggered another one and succeed.

test/simple.test.js
emilgoldsmith: Ahh yes of course, didn't think of that, true!

emilgoldsmith:
Great @chinesedfan ! Sorry for the wait, I kept on getting away from looking at this. It looks great now, so I think I'll just merge it in in 24 hours unless @ismay or @mxstbr have any objections?

emilgoldsmith:
I'll assume that means all good, also as I think it all looks like it should simply be a good improvement!

Thanks again for your contribution @chinesedfan !

PR #146, opened 2017-11-25
Wrap whole transform block in try/catch

asvetliakov:
This will catch errors when filepath is undefined.

screen shot 2017-11-25 at 05 59 43

Happens when viewing diff from git tab in vscode.


emilgoldsmith:
Yep, looks good! Thanks for the PR :)

emilgoldsmith:
Just published `v1.1.1` so you should be able to use it now :)

PR #149, opened 2017-12-05
Fix if no file name given

### LINKED ISSUE #76: Breaks stylelint if no file name given
If you either use the Node API `code` option without using `codeFilename`, or if you lint through `stdin` this will currently make stylelint throw an error with our processor as we require `codeFilename` to be specified for (I believe this is the reason without having looked much more into it anyway) source maps. 

We should probably just quickly handle that case.

chinesedfan:
Fixes #76. The input id matches [the logic of postcss](https://github.com/postcss/postcss/blob/master/lib/input.es6#L78).

Please notice that current `package-lock.json` uses `stylelint@8.0.0`, which doesn't validate the existence of `codeFilename`. But from [8.1.0](https://github.com/stylelint/stylelint/blob/master/CHANGELOG.md#810), non-existed `codeFilename` will also cause `filepath` to be undefined. If you want to upgrade `stylelint`, my second commit can be reverted.

### DIFF
--- a/test/nofiles.test.js
+++ b/test/nofiles.test.js
@@ -62,35 +62,9 @@ describe('no files', () => {
     })
   })
 
-  describe('not existed codeFilename', () => {
+  describe('has codeFilename', () => {
     beforeAll(() => {
-      codeFilename = 'not-existed-file'
-    })
-
-    it('should have one result', () => {
-      expect(data.results.length).toEqual(1)
-    })
-
-    it('should use the right file', () => {
-      expect(data.results[0].source).toEqual('<input css 7>')
-    })
-
-    it('should have errored', () => {
-      expect(data.errored).toEqual(true)
-    })
-
-    it('should have one warnings', () => {
-      expect(data.results[0].warnings.length).toEqual(1)
-    })
-
-    it('should have the right line number', () => {
-      expect(data.results[0].warnings[0].line).toEqual(3)
-    })
-  })
-
-  describe('existed codeFilename', () => {
-    beforeAll(() => {
-      codeFilename = 'package.json'
+      codeFilename = 'somefile.js'
     })
 
     it('should have one result', () => {
--- a/test/nofiles.test.js
+++ b/test/nofiles.test.js
@@ -1,0 +1,116 @@
+const stylelint = require('stylelint')
+const path = require('path')
+
+const code = `import styled from 'styled-components';
+const StyledTable = styled(StyledTableBase)\`
+color: red;
+\``
+const processor = path.join(__dirname, '../lib/index.js')
+const rules = {
+  indentation: 2
+}
+
+describe('no files', () => {
+  let codeFilename
+  let data
+
+  // NOTE beforeEach() runs _after_ the beforeAll() hooks of the describe() blocks
+  beforeEach(done => {
+    stylelint
+      .lint({
+        code,
+        codeFilename,
+        syntax: 'scss',
+        config: {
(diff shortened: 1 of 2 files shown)


test/nofiles.test.js
```diff
@@ -0,0 +1,90 @@
...
+const processor = path.join(__dirname, '../lib/index.js')
+const rules = {
+  indentation: 2
+}
+
+describe('no files', () => {
+  let codeFilename
+  let data
+
+  // NOTE beforeEach() runs _after_ the beforeAll() hooks of the describe() blocks
```
emilgoldsmith: Good comment, I actually would've thought it was the other way since beforeAll is nested inside the next block

test/nofiles.test.js
```diff
@@ -0,0 +1,90 @@
...
+      .catch(err => {
+        // eslint-disable-next-line
+        console.log(err)
+        data = err
+        done()
+      })
+  })
+
+  describe('no codeFilename', () => {
+    beforeAll(() => {})
```
emilgoldsmith: This is redundant right? We can just delete it. Unless you wanna assign `codeFileName = undefined;` specifically, but tests in Jest are guaranteed to run in order right? So I think I'm okay with both, deleting or explicitly defining it to be `undefined`

test/nofiles.test.js
```diff
@@ -0,0 +1,90 @@
...
+
+    it('should have one result', () => {
+      expect(data.results.length).toEqual(1)
+    })
+
+    it('should use the right file', () => {
+      expect(data.results[0].source).toEqual('<input css 2>')
```
emilgoldsmith: This is more curiosity, as the test obviously passes. The reason the id here is 2 is because the `beforeEach` runs before each of these and therefore for each `it` call, a new linting occurs which increments the id?

test/nofiles.test.js
```diff
@@ -0,0 +1,90 @@
...
+    })
+
+    it('should have errored', () => {
+      expect(data.errored).toEqual(true)
+    })
+
+    it('should have one warnings', () => {
```
emilgoldsmith: Very picky, but I guess there shouldn't be an s on the warning, I might have let it go if I hadn't asked for another change though.

test/nofiles.test.js
chinesedfan: Ha, it is not my invention. I imitated from other test files.

test/nofiles.test.js
chinesedfan: Yes, I defined it explicitly.

test/nofiles.test.js
chinesedfan: Exactly as you thought. If it is hard to understand, shall we add comments here?

test/nofiles.test.js
chinesedfan: The more picky, the better. Thanks for pointing out!

test/nofiles.test.js
emilgoldsmith: Yeah I see haha! Too long since I looked at the test files myself I guess! :)

test/nofiles.test.js
emilgoldsmith: Yeah sure, why not? :). And then I'll merge it in right away.

chinesedfan:
@emilgoldsmith I added the comment by amending the last commit directly. So there are still 3 commits.

emilgoldsmith:
Awesome, let's get it live!

PR #150, opened 2017-12-10
Increase async test timeout

emilgoldsmith:
I was just getting sick of all the failed Travis builds which were mostly due to timeouts when doing the coverage which I guess makes the tests take that little longer that makes the tests time out on the relatively slow CI CPU, so added this.

I guess this should just be mergeable, but for good measure if @mxstbr or @ismay could just approve it?


emilgoldsmith:
> Seems weird to have to set it in every test file...

It does indeed! And I actually thought we had to, and have done similarly in other projects, but you motivated me to look deeper, and I finally found a way :)

PR #164, opened 2018-01-22
Add test cases for line comments

### LINKED ISSUE #156: Support JS style line comments (// comment)
For example:
```js
export default styled.div`
  width: 100%; // fails
`;
```

Converting to a typical CSS comment `/* */` avoids the error. But line comments are convenient, and my understanding is that styled-components supports them.

```
├─ stylelint-config-recommended@2.0.0
├─ stylelint-config-standard@18.0.0
├─ stylelint-config-styled-components@0.1.1
├─ stylelint-processor-styled-components@1.2.1
└─ stylelint@8.3.1
```

emilgoldsmith:
Test cases for JS style line comments, this should close #156 either by adding the functionality if we find out it is actually broken or just by merging this PR with the passing tests that properly document that we support this

### DIFF
--- a/test/fixtures/hard/valid-js-comments.js
+++ b/test/fixtures/hard/valid-js-comments.js
@@ -1,0 +1,10 @@
+import styled from 'styled-components';
+
+const Button1 = styled.div`
+  color: red;
+  display: block;
+  // Some comment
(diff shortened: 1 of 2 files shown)


emilgoldsmith:
I'm merging this in as it's pretty simple and has been here for a while

PR #167, opened 2018-02-21
Upgrade Stylelint which enables supporting sourcemaps on css syntax errors

### LINKED ISSUE #154: Source maps don't get assigned during CssSyntaxError
It seems from #153 that the source maps don't run, and actually not even the `result()` function in `index.js` is run when there's a CssSyntaxError, I'm not quite sure if this is something we can actually do something about as it seems that Stylelint simply doesn't "give back control" to us in the case of a CssSyntaxError, but it should definitely be looked into as it's of course confusing behavior for users.

emilgoldsmith:
closes #166 and resolves #154.

A new version of Stylelint was finally published which includes my feature I added to Stylelint so they now let us apply sourcemaps to CssSyntaxErrors, I modified our code slightly to now act properly in that case and added a few tests.


emilgoldsmith:
Just upgraded to 9.1.1 as that was just released,

Also added Stylelint v9+ as a PeerDependency as we need this for all of our functionality to work (and for npm versions for the last many years it doesn't break installation just produces a warning and I think that's okay).

and @chinesedfan I tried reverting your commit but it's failing the test, it doesn't act as you expected, you just want me to undo that revert or if you want I'm not planning on touching this branch anymore right now, so feel free to just go in and change whatever you want to make it work

emilgoldsmith:
@ismay fair enough!

chinesedfan:
> But from 8.1.0, non-existed codeFilename will also cause `filepath` to be undefined.

@emilgoldsmith You are right. We can't simply revert that commit. I thought I mistook `filepath` in `stylelint/lib/createStylelintResult.js` with the one in our processor.

I have undo the revert and the corresponding typo fix. Before we merge this PR, dropping them totally will make the branch more clear.

PR #84, opened 2017-08-11
Fix interpolation following a comment edge case

emilgoldsmith:
This resolves #79 

This is done and ready for review.

PR #174, opened 2018-03-09
Fixed interpolations following a multiline comment

philipvonbargen:
The previous PR #84 fixed comments before interpolations but does not work for multiline comments.

The updated RegEx will now match any characters that are not `*/` (https://regexr.com/3m0r1).

```css
/* single line */
display: block;

/* 
  multiline with "*" and "/"
*/
display: block;


/** 
 * JsDoc style
 */
display: block;
```

### DIFF
--- a/src/utils/general.js
+++ b/src/utils/general.js
@@ -67,7 +67,7 @@ const reverseString = str => str.split('').reverse().join('')
 
 const isLastDeclarationCompleted = text => {
   // We disregard all comments in this assessment of declaration completion
-  const commentsRemoved = text.replace(/\/\*.*?\*\//g, '')
+  const commentsRemoved = text.replace(/\/\*[\s\S]*?\*\//g, '')
   const reversedText = reverseString(commentsRemoved)
   const lastNonWhitespaceChar = nextNonWhitespaceChar(reversedText)
   if (
--- a/test/fixtures/interpolations/valid.js
+++ b/test/fixtures/interpolations/valid.js
@@ -116,3 +116,19 @@ const Button12 = styled.button`
   /* stylelint-disable */
   ${colorExpression}
 `
+
+const Button13 = styled.button`
+  display: block;
+  /*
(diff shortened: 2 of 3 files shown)


src/utils/general.js
```diff
@@ -67,7 +67,7 @@ const reverseString = str => str.split('').reverse().join('')
 
 const isLastDeclarationCompleted = text => {
   // We disregard all comments in this assessment of declaration completion
-  const commentsRemoved = text.replace(/\/\*.*?\*\//g, '')
+  const commentsRemoved = text.replace(/\/\*[\s\S]*?(?!\*\/)?\*\//g, '')
```
emilgoldsmith: What is this supposed to do `(?!\*\/)?`? Is it supposed to be a conditional negative lookahead? Doesn't that just match everything then? I also don't think you need it, the `*?` is a non-greedy matcher so it'll stop as soon as it can

philipvonbargen: I played around a bit more in regexr and managed to shorten it a little to `/\/\*(?!\*\/)[\s\S]*?\*\//g`. This is doing a negative lookahead of the sequence `*/` and then lazily matches any other characters. Without the `?` it would match everything between multiple comment blocks.

mxstbr:
Thank you so much for helping us improve styled-components! Based on our [Community Guidelines](https://github.com/styled-components/styled-components/blob/master/CONTRIBUTING.md) every person that has a PR of any kind merged is offered **an invitation to the styled-components organization**—that is you right now!

Should you accept, you'll get write access to the main repository. (and a fancy `styled-components` logo on your GitHub profile!) You'll be able to label and close issues, merge pull requests, create new branches, etc. We want you to help us out with those things, so don't be scared to do them! Make sure to read our [contribution and community guidelines](https://github.com/styled-components/styled-components/blob/master/CONTRIBUTING.md), which explains all of this in more detail. Of course, if you have any questions just let me know!

PR #190, opened 2018-05-24
Remove extra npm install

chinesedfan:
Usually, `npm` is installed with `nodejs`. And latest npm@6.1.0 is not compatible with nodejs@4, which will break our CI.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,7 @@
   "presets": [
     ["env", {
       "targets": {
-        "node": 4
+        "node": 6
       }
     }]
   ]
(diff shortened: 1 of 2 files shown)


chinesedfan:
Interesting. npm@2 will fail with unmet peer dependencies. And the greenkeeper has stopped working to upgrade stylelint to fix this problem.

emilgoldsmith:
Hmm, node v4 is not LTS anymore either though, and I know many major packages have stopped supporting it, so maybe we should also just remove it from our CI?

chinesedfan:
Sure, I can revert those nodejs@4 related commits this weekend.

emilgoldsmith:
Awesome, thanks!

PR #191, opened 2018-05-24
Remove installing latest npm in CI

emilgoldsmith:
It's making node v4 fail because I guess the latest version of npm isn't compatible with it, and I think it should also be okay using the version of npm that comes with each node version in the CI?


emilgoldsmith:
Whoops duplicate of #190

Issue #170, opened 2018-03-03
Remove `syntax: 'scss'` from config in docs and tests


jeddy3:
Firstly, thank you for creating and maintaining this processor. I use it myself when working with styled-components.

I had a query about the `syntax: scss` property value pair in the [Setup section](https://github.com/styled-components/stylelint-processor-styled-components#setup) of the README. 

```js
{
  "processors": ["stylelint-processor-styled-components"],
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-styled-components"
  ],
  "syntax": "scss"
}
```

Is that pair unique to this processor as I don't believe it is used in stylelint itself? I did a cursory scan of the code base here, but couldn't find where it was used.

emilgoldsmith:
It isn't, but it's probably infrequently used because Stylelint normally infers it from file extension, this is the closest I found to documentation on it from a quick look https://stylelint.io/user-guide/css-processors/#parsing-non-standard-syntax and here https://stylelint.io/user-guide/faq/#how-do-i-lint-scss-less-or-other-non-standard-syntax

jeddy3:
Thank you for the reply. 

> it's probably infrequently used

There was a request in https://github.com/stylelint/stylelint/issues/2973 to add support for setting the syntax within the configuration object, but we found an alternative solution and the feature was never implemented. So, I don't believe stylelint will respect this configuration object property.

I also tried not using the property alongside nesting and interpolation on a local project, and it seemed fine:

```js
// src/components/TextControl/index.js
import React from "react";
import styled from "styled-components";

const Feedback = styled.div`
  display: none;
  ${/*sc-selector*/ Input}:focus ~ & {
    background-color: red;
    bottom: -16px;
  }
`;
```

```js
{
  "processors": ["stylelint-processor-styled-components"],
  "rules": {
      "selector-combinator-whitelist": []
  }
}
```

```shell
jeddy3:bds jeddy3$ ./node_modules/.bin/stylelint "src/**/*.js"

src/components/TextControl/index.js
  6:13  ✖  Unexpected combinator "~"   selector-combinator-whitelist
```

Would you like me to submit a PR to update the documentation?

emilgoldsmith:
Interesting, @mxstbr wanna chime in? I guess it's you that set this requirement originally?

mxstbr:
Yeah I thought it was necessary, nevermind then! If it works let's remove that from the docs

emilgoldsmith:
[...]

PR #197, opened 2018-05-31
Fix JS style comments are not tested

chinesedfan:
Be careful. Without corresponding rules set, the test will be always passed.

### DIFF
--- a/test/hard.test.js
+++ b/test/hard.test.js
@@ -4,6 +4,7 @@ const path = require('path')
 const processor = path.join(__dirname, '../lib/index.js')
 const rules = {
   'block-no-empty': true,
+  'property-no-unknown': true,
   indentation: 2
 }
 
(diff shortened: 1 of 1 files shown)


test/hard.test.js
```diff
@@ -17,6 +18,7 @@ describe('hard', () => {
     stylelint
       .lint({
         files: [fixture],
+        syntax: 'scss',
```
emilgoldsmith: We removed this in #170 as it isn't actually a stylelint feature apparently

chinesedfan: No, it is only not in `.stylelintrc` as #170 did, but it exists in [Node.js API](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/node-api.md#syntax).

emilgoldsmith: Ahh okay

Issue #210, opened 2018-08-24
Typed parameters cause file to be ignored


ubnt-marc-khouri:
When using typed parameters for styled-components in Typescript, stylelint skips the file and reports no output at all for it. Note that the method of typing illustrated before is slightly different than the one used in [your typescript test](https://github.com/styled-components/stylelint-processor-styled-components/blob/master/test/fixtures/typescript/ts-syntax-valid.ts#L14). I'm not certain how to gather further debug information for this.

Minimal repro (filename: `test.ts`):

```typescript
import styled from 'styled-components';

// Stylelint usually catches the missing semi on 'width'
const LinterRuns = styled.div`
  width: ${p => p.width || '100%'}
  height: '100%';
`;

// The code segment below breaks stylelint. No errors
// will be reported for the file. If you comment out everything
// below this point, stylelint will catch the missing semi above
const LinterIsBroken = styled.div<{ width?: string }>`
  width: ${p => p.width || '100%'}
  height: '100%';
`;
```

command:

```
⇒  ./node_modules/.bin/stylelint './test.ts'
# returns exit code 0 for the file above
```

.stylelintrc:

```
{
  "processors": ["stylelint-processor-styled-components"],
  "extends": [
    "stylelint-config-recommended",
    "stylelint-config-styled-components"
  ]
}
```

versions:
- stylelint@9.5.0
- stylelint-config-recommended@2.1.0
- stylelint-config-styled-components@0.1.1
- stylelint-processor-styled-components@1.3.2
- styled-components@3.3.3
- typescript@3.0.1
- node v8.11.3

chinesedfan:
Very detailed issue! It should be solved by #163, but was hanging there so long. I will try to merge it as soon as possible.

If the processor throws out the real error, we can see following messages.

```
SyntaxError: Unexpected token, expected "," (12:41)
    at _class.raise (/Users/zhong/Desktop/opensource/stylelint-processor-styled-components/node_modules/babylon/lib/index.js:776:15)
    at _class.unexpected (/Users/zhong/Desktop/opensource/stylelint-processor-styled-components/node_modules/babylon/lib/index.js:2079:16)
    at _class.expect (/Users/zhong/Desktop/opensource/stylelint-processor-styled-components/node_modules/babylon/lib/index.js:2067:28)
```

brettjurgens:
[...]

Issue #204, opened 2018-07-19
Missing semicolon not detected


rylek90:
I use `stylelint-processor-styled-components` with Emotion. With current update Emotion doesn't automatically add semicolons at the end of expression. However missing semicolons are not detected by plugin

My stylelint config:
```
module.exports = {
  processors: ['stylelint-processor-styled-components'],
  extends: [
    'stylelint-config-recommended',
    'stylelint-config-styled-components'
  ],
  rules: {
    'no-descending-specificity': null
    'property-no-vendor-prefix': null,
    'value-no-vendor-prefix': null,
    'no-invalid-double-slash-comments': null,
    'comment-empty-line-before': null,
    'declaration-block-trailing-semicolon': 'always',
    'no-duplicate-selectors': null,
    'property-no-unknown': null,
    'color-named': 'never',
    'color-no-hex': true,
    'function-blacklist': ['rgb', 'rgba']
  }
}
```

Linted code (with missing semicolons):
```
import styled from 'react-emotion'

export const MyFancyButton = styled(Button)`
 padding: none
 color: inherit
 border: none
`
```

I use `stylelint --syntax scss`

chinesedfan:
Hi @rylek90, which stylelint rule do you think will be violated if semicolons are missing?

rylek90:
Hmm, that would be a simple CssSyntaxError I believe 🤔

chinesedfan:
Yes, I tested with a simple code snippet and got CssSyntaxError. Can you give detailed versions of your stylelint and related packages? Or the link of the problem project directly?

rylek90:
@chinesedfan versions are following:
```
"stylelint": "9.3.0",
"stylelint-config-standard": "18.2.0",
"stylelint-config-styled-components": "0.1.1",
"stylelint-processor-styled-components": "1.3.1"
```

chinesedfan:
@rylek90 I still got CssSyntaxError. What is your expectation?

By the way, you set `stylelint-config-recommended` in stylelint configuration, but why is `stylelint-config-standard` used?

rylek90:
@chinesedfan my apologise, I've spot the error finally. It seems like it has problems with js decorators. Here is a minimal repo to reproduce https://github.com/rylek90/StyledComponentsBug

chinesedfan:
@rylek90 I see. It is due to the babylon error isn't be thrown by our processor. #163 has tried to handle it but not finished yet.

```
[...]

PR #211, opened 2018-08-25
Wrap non-CssSyntaxError into stylelint results

chinesedfan:
At the beginning, this processor will throw every error to stylelint. But stylelint can only handle `CssSyntaxError`. (#131)

After #142, non-CssSyntaxErrors are ignored, which causes some confusions(#204, #210).

I think we should wrap those errors into stylelint results. So that both requirements are met.


src/index.js
```diff
@@ -38,22 +38,32 @@ module.exports = options => ({
...
+    const err = errorWasThrown[filepath]
+    if (err) {
+      if (err.name === 'CssSyntaxError') {
+        // We threw an error ourselves, in this case we have already put correct
+        // line/column numbers so no source maps are needed
+        // (and would actually break the line numbers)
+        return stylelintResult
+      } else {
+        // For other errors, wrap them into the result
+        return Object.assign({}, stylelintResult, {
```
emilgoldsmith: Does this get displayed to the user?

Issue #183, opened 2018-05-15
styled-components with stylelint & babel-proposal-optional-chaining


chrishelgert:
I am trying to use styled-components with stylelint and babel-proposal-optional-chaining.
As soon as a styled-components uses optional-chaining, stylelint will skip the whole file.

```javascript
// package.json
"stylelint": "9.1.3",
"stylelint-config-standard": "^18.2.0",
"stylelint-config-styled-components": "^0.1.1",
"stylelint-processor-styled-components": "^1.3.1"
```

```javascript
// .stylelintrc
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-styled-components"
  ],
  "processors": ["stylelint-processor-styled-components"],
  "syntax": "scss"
}
```

*Will not be linted:*
```javascript
const button = styled.button`
  background-color: ${backgroundColor};
  border-color: ${props => props.theme.button?.borderColor || 'transparent'};
  border-style: solid;
  border-width: ${props => props.theme.button?.borderWidth || '1px'};
  border-radius: ${props => props.theme.button?.borderRadius || '0.25rem'};
  color: ${props => props.theme.button?.color || '#fff'};
  cursor: pointer;
  font-family: ${props => props.theme.fonts.primary};
  font-size: ${props => props.theme.button?.fontSize || '1rem'};
  line-height: ${props => props.theme.button?.lineHeight || '1.5'};
  padding: ${props => props.theme.button?.padding || '0.375rem 0.75rem'};
  width: ${props => props.theme.button?.width || '100%'};
  &:hover {
    background-color: ${backgroundColorHover};
  }
  ${props => props.theme.button?.custom}
`
```

*Will be linted:*
```javascript
const Container = styled.div`
  display: flex;
  margin-bottom: 30px;
`
```

*Demo Test:*
```javascript
// fixture
import styled from 'styled-components'

export default styled.button`
  background-color: ${backgroundColor};
  border-color: ${props => props.theme.button?.borderColor || 'transparent'};
  border-style: solid // there is the error
  border-width: ${props => props.theme.button?.borderWidth || '1px'};
  border-radius: ${props => props.theme.button?.borderRadius || '0.25rem'};
  color: ${props => props.theme.button?.color || '#fff'};
  cursor: pointer;
  font-family: ${props => props.theme.fonts.primary};
  font-size: ${props => props.theme.button?.fontSize || '1rem'};
[...]

PR #229, opened 2018-10-02
Support to specify parser plugins

### LINKED ISSUE #228: `exportDefaultFrom` error from parser after upgrade to 1.4.0
I'm currently in the process of updating from version `1.3.2` to `1.4.0`. As I can see from the changelog, babel errors are now shown to users. I'm now getting a few of those:

    This experimental syntax requires enabling the parser plugin: 'exportDefaultFrom' (1:7)

The files where this occurs are very simple e.g.:

    export Container from './Container';

How can I configure the parser used by this plugin to allow for the experimental syntax? Am I doing something wrong?

chinesedfan:
Fixes #228 and solved #183 in another way.

Now users can use `options.parserPlugins` to custom plugins to babel.


rylek90:
Is there an option not to rewrite whole array but rather concatenate with default one?
Can we work towards passing whole Babel configuration in the future?

chinesedfan:
@rylek90 See #231. Please feel free to send a PR if you would like do that.

rylek90:
@chinesedfan thanks, I will take a look at in in the free time 🙌

chinesedfan:
@jeffcatania I am afraid not because #231 should be the better way. You can use the test file([options.test.js#L211](https://github.com/styled-components/stylelint-processor-styled-components/blob/master/test/options.test.js#L211)) as a Node API example now. Following links may 
 also help you.

- How to config processor options: https://stylelint.io/user-guide/configuration/#processors
- Available babel parser plugins: https://babeljs.io/docs/en/babel-parser#plugins

PR #243, opened 2018-11-26
make install script copy paste friendly

martonlanga:


chinesedfan:
@martonlanga Sorry, I think the original script is easier to be read and no problem to be executed. Did you forget to copy the parentheses？

martonlanga:
Aah it works like that. Cool then this format is much better! Cheers

chinesedfan:
Good. So can I close this PR, right? Thanks for your patch all the same.

Issue #231, opened 2018-10-05
Support common ways to config babel


chinesedfan:
As suggested in #228, hope this processor read babel configurations from `.babelrc` or `package.json`.

PR #252, opened 2019-03-05
Update babylon-parser.js (Add nullishCoalescingOperator)

kaywyeex:
Nullish coalescing, like optional chaining is at stage 1.

We've added our own parserPlugins to processor options, but now we have to maintain ourselves.
This problem is easily solved by adding it to the default plugin list. 🌮

### DIFF
--- a/src/parsers/babylon-parser.js
+++ b/src/parsers/babylon-parser.js
@@ -15,7 +15,8 @@ module.exports = (type, plugins) => input =>
         'functionSent',
         'dynamicImport',
         'optionalCatchBinding',
-        'optionalChaining'
+        'optionalChaining',
+        'nullishCoalescingOperator'
       ]
     )
   })


kaywyeex:
Bump for justice. 🌮

chinesedfan:
Yes, you have to maintain yourselves. If added too much plugins by default, users who don't need them will have slower linting unnecessarily. It looks like a little unfair. But optional chaining was added before `options.parserPlugins` was supported.

We hope #231 can become a better solution. Let's share ideas there.

fabb:
Nullish coalescing is now at stage 3 and soon stage 4, it should be added to the default plugin list

dimitropoulos:
for anyone hitting this issue that would like explicit instructions on the temporary resolution of this problem (i.e. temporary until this project adds the rule themselves since this is in the language now), please see here: https://github.com/stylelint/stylelint/issues/4473

PR #258, opened 2019-03-25
add `strict` option

### LINKED ISSUE #253: Unable to specify a specific import from a specific module
So you provide two options:

* `moduleName`
* `importName`

Used like so:

```ts
// moduleName = module
import {x, y} from 'module';
x`test`; // parsed
y`test`; // parsed
```

```ts
// moduleName = module
// importName = y
import {x, y} from 'module';
x`test`; // parsed
y`test`; // parsed
y(whatever)`test`; // parsed
y.whatever`test`; // parsed
```

Surely this isn't how `importName` should work. I would expect only tagged templates which use the specified import name to be parsed, rather than ALL imported names _and_ the extra styled-specific stuff.

Like this:

```ts
// moduleName = module
// importName = y
import {x, y} from 'module';
x`test`; // NOT parsed
y`test`; // parsed
```

With this ability we could use the same component on other tagged template css libraries. Something like [LitElement](https://github.com/Polymer/lit-element/). But we can't yet because we over-parse:

```ts
css`this is css`; // parsed
html`this is html`; // parsed <- WRONG, it shouldn't be
```

43081j:
Fixes #253.

Enables stricter parsing, in that only expressions using the specified
`importName` from the `moduleName` will be parsed.

```ts
// moduleName = some-module, importName = foo
import { foo, bar } from 'some-module';
foo`test`; // parsed
bar`test`; // parsed
foo.bleh`test`; // parsed
bar.bleh`test`; // parsed

// moduleName = some-module, importName = foo, strict = true
import { foo, bar } from 'some-module';
foo`test`; // parsed
bar`test`; // ignored
foo.bleh`test`; // parsed
bar.bleh`test`; // ignored
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,7 +13,8 @@ const sourceMapsCorrections = {}
 const errorWasThrown = {}
 const DEFAULT_OPTIONS = {
   moduleName: 'styled-components',
-  importName: 'default'
+  importName: 'default',
+  strict: false
 }
 
 module.exports = options => ({
--- a/src/parsers/index.js
+++ b/src/parsers/index.js
@@ -44,7 +44,9 @@ const processStyledComponentsFile = (ast, absolutePath, options) => {
         importedNames = parseImports(node)
         return
       }
-      const helper = isHelper(node, importedNames)
+      const helper = !options.strict
+        ? isHelper(node, importedNames)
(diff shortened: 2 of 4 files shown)


chinesedfan:
@43081j Just released 1.6.0. Thanks for your efforts and enjoy!

Issue #187, opened 2018-05-21
Unable to lint CSS and Styled Components with processor enabled


stramel:
Using `v9.2.1` I am able to lint styled-components once again, however I am not able to lint CSS files while the processor is enabled.

emilgoldsmith:
This was never possible, you will need to use a different Stylelint config that doesn't include our processor in order to lint CSS files

stramel:
Dang, that makes it difficult for handling stylelint through vscode.

emilgoldsmith:
Yeah, it sucks. If you want to there might be ways around it, if we could somehow detect whether a file is purely CSS and not JS, such as possibly look at the extension of a file, or use a library that can look at the statics of the code and check whether it parses correctly as CSS and then in that case not apply our processor to it but just pass it through to Stylelint as is. In that case it could probably work. Feel free to work on it and see if you can make it work

emilgoldsmith:
If you want to try working on it feel free to ask any questions if you encounter any problems

brandonkal:
I'm not even sure how I would begin.  This is pretty complex because even if this processor could be bypassed when a CSS file is detected there is still the matter of getting stylelint to ignore the "stylelint-config-styled-components" extension. I am not sure if that is possible.

Right now I have it set up this way: most of my plain CSS is in one directory with its own .stylelintrc file.

In the package.json:
```
"lint:css": "stylelint 'src/**/*.{js,jsx}' && stylelint 'src/**/*.css' --config extras/.stylelintrc"
```

The main .stylelintrc file in the src root directory:

```
{
  "processors": ["stylelint-processor-styled-components"],
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-prettier",
    "stylelint-config-styled-components"
  ]
}
```

And then inside the extras folder:
```
# This file is used by the stylelint CLI to lint CSS files.
# The .stylelintrc file in the project root directory is configured to lint
# CSS in JS within Styled Components
{
  "extends": [
    "stylelint-config-standard",
    "stylelint-config-prettier",
  ]
}
```

This works pretty well. Atom-lint will highlight CSS lint errors in the `extras` folder and JS files in the `src` directory. For all other CSS files, the package script must be run in the console.

emilgoldsmith:
[...]

PR #267, opened 2019-05-22
Add option ignoreFiles to filter fileTypes

schoenwaldnils:
Adding option `ignoreFiles` to filter file types.
In my case this way I can filter `*.css` files to use the processor only in `js, jsx` files.

Solves #187.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 const path = require('path')
+const micromatch = require('micromatch')
 const parse = require('./parsers/index')
 const isCausedBySubstitution = require('./utils/result').isCausedBySubstitution
 const getCorrectColumn = require('./utils/result').getCorrectColumn
@@ -14,10 +15,11 @@ const errorWasThrown = {}
 const DEFAULT_OPTIONS = {
   moduleName: 'styled-components',
   importName: 'default',
-  strict: false
+  strict: false,
+  ignoreFiles: []
 }
 
-module.exports = options => ({
+const realProcessor = options => ({
   // Get string for stylelint to lint
   code(input, filepath) {
     let absolutePath
@@ -28,11 +30,15 @@ module.exports = options => ({
       inputId += 1
     }
 
+    const fileIsIgnored = micromatch(filepath, options.ignoreFiles).length
+
+    if (fileIsIgnored) return input
+
     try {
       const { extractedCSS, interpolationLines, taggedTemplateLocs, sourceMap } = parse(
         input,
         absolutePath,
-        Object.assign({}, DEFAULT_OPTIONS, options)
+        options
       )
       // Save `loc` of template literals
       taggedTemplateLocsMap[absolutePath] = taggedTemplateLocs
@@ -58,6 +64,10 @@ module.exports = options => ({
   // Fix sourcemaps
   result(stylelintResult, filepath) {
     const err = errorWasThrown[filepath]
+    const fileIsIgnored = micromatch(filepath, options.ignoreFiles).length
+
+    if (fileIsIgnored) return stylelintResult
+
     if (err) {
       if (err.name === 'CssSyntaxError') {
         // We threw an error ourselves, in this case we have already put correct
@@ -111,3 +121,5 @@ module.exports = options => ({
     return result
   }
 })
+
+module.exports = options => realProcessor(Object.assign({}, DEFAULT_OPTIONS, options))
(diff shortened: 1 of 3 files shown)


emilgoldsmith:
Hi, it seems you are failing some unit tests at the moment? Could you please make sure those pass, and then someone will try and get to reviewing this :)

src/index.js
```diff
@@ -14,7 +15,8 @@ const errorWasThrown = {}
 const DEFAULT_OPTIONS = {
   moduleName: 'styled-components',
   importName: 'default',
-  strict: false
+  strict: false,
+  ignoreFiles: []
 }
 
 module.exports = options => ({
```
chinesedfan: Tests fail due to `options` is not normalized. A simple adjustment can be,

```
const processerFactory = options => ({
 ... ...
});
module.exports = (options) => processerFactory(Object.assign({}, DEFAULT_OPTIONS, options));
```

src/index.js
```diff
@@ -83,25 +87,31 @@ module.exports = options => ({
...
-            lineCorrection[warning.line] || warning.line,
-            warning.column
-          )
-        })
-      )
+
+    let warnings = stylelintResult.warnings
+
+    const absolutePath = path.resolve(process.cwd(), filepath)
+    if (!micromatch(absolutePath, options.ignoreFiles).length) {
```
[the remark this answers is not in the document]
schoenwaldnils: Like this?
```
result(stylelintResult, filepath) {
  const err = errorWasThrown[filepath]
  const absolutePath = path.resolve(process.cwd(), filepath)
  const fileIsIgnored = micromatch(absolutePath, options.ignoreFiles).length

  if (fileIsIgnored) return stylelintResult
```

src/index.js
schoenwaldnils: I'm not sure what you mean by the `processerFactory` function.
But I mapped the options now and the tests are green

chinesedfan: `processerFactory` means the whole original codes, and please forgive my typo. It should be `processorFactory` or `realProcessor`. If writing in this way, you can handle options in the unique place.

```
const realProcessor = options => ({
  code() {},
  result() {}
});
module.exports = (options) => realProcessor(Object.assign({}, DEFAULT_OPTIONS, options));
```

src/index.js
```diff
@@ -83,7 +96,10 @@ module.exports = options => ({
     const taggedTemplateLocs = taggedTemplateLocsMap[filepath] || []
     const interpolationLines = interpolationLinesMap[filepath] || []
     const lineCorrection = sourceMapsCorrections[filepath]
-    const warnings = stylelintResult.warnings
+
+    let warnings = stylelintResult.warnings
+
+    warnings = stylelintResult.warnings
```
chinesedfan: Could you recover these unnecessary changes?

src/index.js
```diff
@@ -58,6 +65,12 @@ module.exports = options => ({
   // Fix sourcemaps
   result(stylelintResult, filepath) {
     const err = errorWasThrown[filepath]
+    const mappedOptions = Object.assign({}, DEFAULT_OPTIONS, options)
+    const absolutePath = path.resolve(process.cwd(), filepath)
```
chinesedfan: Maybe `filepath` here is the absolute path already, or those `taggedTemplateLocs/interpolationLines` can't get value in old codes.

src/index.js
schoenwaldnils: Done

src/index.js
schoenwaldnils: Compared them. Were the same with the files I tested.
Therefore removed the `absolutePath`

src/index.js
chinesedfan: @schoenwaldnils Looks like we are almost there. Would you mind processing options like above and squashing commits to less than or equal to 3? Thanks for your work in advance.

schoenwaldnils: Now I understood what you meant with the processing ^^
Done

tomdev10:
Hi. Is there any documentation for this? The syntax @schoenwaldnils suggests in https://github.com/styled-components/stylelint-processor-styled-components/issues/187#issuecomment-494827256 doesn't seem to work for me. Cheers

chinesedfan:
@tomdev10 Check #272. Notice items in config [`processors`](https://stylelint.io/user-guide/configuration#processors) are still arrays.

PR #275, opened 2019-08-20
updated dependencies @babel/parser and @babel/traverse

### LINKED ISSUE #274: Upgrade outdated "@babel/parser" dependency
This lib still depends on old versions of the @babel/parser and @babel/traverse. This makes this plugin incompatible with newer typescript syntax.

```
  "dependencies": {
    "@babel/parser": "^7.0.0",
    "@babel/traverse": "^7.0.0",
```

https://github.com/styled-components/stylelint-processor-styled-components/blob/master/package.json#L64,L65

For example when using `as const` TypeScript syntax this causes a problem:

```
const positionFixedOrAbsoulteDecoder = JsonDecoder.oneOf(
    (['fixed', 'absolute'] as const).map(s => JsonDecoder.isExactly(s)),
    'position fixed/absolute'
)
```

```
 99:30  ✖  Unexpected token (99:30)
```

The fix is to update these dependencies to >= 7.4

fabb:
This enables support for the latest typescript syntax (e.g. `as const`).

Fixes #274.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,6 @@
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
       "integrity": "sha1-BuKrGb21NThVWaq7W6WXKUgoAPg=",
-      "dev": true,
       "requires": {
         "@babel/highlight": "^7.0.0"
       }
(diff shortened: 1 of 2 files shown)


chinesedfan:
@fabb Thanks for your PR. According to https://github.com/styled-components/stylelint-processor-styled-components/issues/274#issuecomment-523025622, it can be unmodified.

PR #272, opened 2019-07-16
document options

hipstersmoothie:
I didn't see the options documented anywhere and had to find them through searching issues and reading code. This PR tries to use the original authors words to describe the features

PR #279, opened 2019-08-30
Add docs for options and fix something

chinesedfan:
Pushed by #272.


chinesedfan:
@hipstersmoothie Thanks for your efforts in #272, too.

PR #284, opened 2019-11-10
Add ?? support to default parser plugins

ProdigySim:
I had to add this to my proejct's stylelint config to get my stylelint working after adopting nullish coalescing in my codebase.

Now that Nullish Coalesce is stage 3, and it's enabled by default in projects like Typescript, this might be a useful default to add. Since optional chaining support is already in, this shouldn't be a too radical change. Just a proposal!

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ And it also has some options. Their default values are,
       "functionBind",
       "functionSent",
       "dynamicImport",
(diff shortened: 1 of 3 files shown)


chinesedfan:
Closed in favor of #295 and 1.10.0 has been published.

PR #289, opened 2019-11-28
added nullishCoalescingOperator to parsers

fabb:
nullish coalescing is now at stage 3 and soon stage 4, therefore it should be added to the parser list
#252

### DIFF
--- a/src/parsers/babylon-parser.js
+++ b/src/parsers/babylon-parser.js
@@ -15,7 +15,8 @@ module.exports = (type, plugins) => input =>
         'functionSent',
         'dynamicImport',
         'optionalCatchBinding',
-        'optionalChaining'
+        'optionalChaining',
+        'nullishCoalescingOperator'
       ]
     )
   })


dimitropoulos:
for anyone hitting this issue that would like explicit instructions on the temporary resolution of this problem (i.e. temporary until this project adds the rule themselves since this is in the language now), please see here: https://github.com/stylelint/stylelint/issues/4473

chinesedfan:
@fabb @ProdigySim Thanks for your fixes(#284, #289). Updating default plugins is the quickest way to enable those new features, but as our [F.A.Q](https://github.com/styled-components/stylelint-processor-styled-components#why-does-it-throw-unexpected-token-even-thought-the-file-didnt-import-styled-components) has mentioned, we hope someone can take a look at #231. That can make users reuse their babel configs, instead of adding more and more plugins in this processor.

chinesedfan:
Closed in favor of #295 and 1.10.0 has been published.

PR #290, opened 2020-01-28
chore: bump dependencies

KubaJastrz:
Changes:
- upgraded all `dependencies` and `devDependencies`
- updated configuration for `jest`, `eslint`, `prettier`, `husky` and `lint-staged`
- linted all files
- wrapped paths with `slash` to fix tests failing on Windows after stylelint@13 breaking changes
- removed babel plugins from `@babel/parser` config that are already supported (in latest 7.8.3 version)

This has been done in preparations for #231


chinesedfan:
@KubaJastrz Really appreciate your work. I almost want to merge it directly except for a few questions.

> wrapped paths with slash to fix tests failing on Windows after stylelint@13 breaking changes

Can you elaborate what breaking changes are?

> removed babel plugins from @babel/parser config that are already supported (in latest 7.8.3 version)

Do you mean those plugins have been added by default? And what's your reason about adding `funciton-sent`?

> This has been done in preparations for #231

Would you mind explaining your plan a little more? Just want to confirm in advance and to avoid changes in vain.

KubaJastrz:
> Can you elaborate what breaking changes are?

It seems like https://github.com/stylelint/stylelint/pull/4254 prevents us from passing fixture file paths directly into `stylelint.lint({ files: [fixture] })` config on Windows, because it uses backward slashes, hence this change:

```diff
- fixture = path.join(__dirname, './fixtures/...')
+ fixture = slash(path.join(__dirname, './fixtures/...'))
```

Instead of `slash` library we could use `path.posix.join`, but then all tests like this would fail because `stylelint` returns back original paths:

```js
it('should use the right file', () => {
  expect(data.results[0].source).toEqual(fixture)
})
```

> Do you mean those plugins have been added by default? And what's your reason about adding `function-sent`?

Yes, for example @babel/parser supports `nullishCoalescingOperator`, `optionalChaining` and `dynamicImport` since [7.8.0](https://babeljs.io/blog/2020/01/11/7.8.0), `objectRestSpread` were `asyncGenerators` already supported in ES2018: https://github.com/babel/babel/pull/8448

I didn't add `function-sent`, it just appears green because of the removed comma `,`.

> Would you mind explaining your plan a little more? Just want to confirm in advance and to avoid changes in vain.

I would like to look at #231, but I wanted to make sure that we are using latest @babel/core first, which added support for [new configuration files](https://babeljs.io/blog/2020/01/11/7.8.0#support-every-configuration-file-extension-10783-https-githubcom-babel-babel-pull-10783-and-10903-https-githubcom-babel-babel-pull-10903) that could come in handy.

